PR #2, opened 2020-07-23
fix github and twitter social links

TortleWortle:
changed a couple of Twitter and Github links from /codenameglimesh to /glimesh


clone1018:
Thanks for fixing these!

PR #3, opened 2020-07-23
Updated README with instructions on cloning

omn0mn0m:
Included `--recursive` in case someone is not familiar with submodules.


clone1018:
Thanks for the docs!

PR #4, opened 2020-07-23
Added a few common footer links

omn0mn0m:
Not a big deal, do with it what you will :)


clone1018:
I want to get these populated before I merge this

clone1018:
Going to probably end up going with an email or form submit route, instead of requiring users to have a GitHub account.

Thank you for the PR!

PR #33, opened 2020-07-24
Fixes #25

### LINKED ISSUE #25: Usernames are case sensitive
Currently https://glimesh.tv/ClOnE1018/profile does not resolve my profile. We'll probably need to either slugify all usernames, or ensure `where` statement searches independent of the case.

mja00:
Adds a new display name field, this field allows for certain capitalization and such. The username field is now used as a slug for various things. All URL requests convert the request to lowercase before querying the DB so you can mix case a username in the url and it will return the correct one. Also automatically converts old users over to the new format.

I merged my branch that fixed the edit profile page into this one so I'll be closing that pull request since this one will now handle it.


mja00:
https://github.com/Glimesh/glimesh.tv/blob/14981b25165fce6bc1c2d28df3e15d127605512e/lib/glimesh/accounts/user.ex#L35-L44

This section is currently causing most tests to fail. It has something due to the test sessions being registered. I plan on taking a look later tonight but if anyone wants to take a crack at it just checkout the branch tied to this PR.

mja00:
I fixed it. :)

clone1018:
I really think we need to have this functionality, but I'm not confident in the changes of the `attrs` structure that caused the tests to have to be changed from atoms to strings.

http://learningwithjb.com/posts/the-many-different-ways-to-key-elixir-maps

I think this line changes the attrs to string based keys, and not atoms, I'm not comfortable with the implications of this: https://github.com/Glimesh/glimesh.tv/pull/33/files#diff-3aa0b5a00c41c9172c107f8f3bd78dceR99

clone1018:
I was able to get similar functionality working over on the blog: https://github.com/Glimesh/glimesh.tv/pull/43/files#diff-c443fd987c6cd31140e5bbb212e9cfddR27

lib/glimesh/accounts/user.ex
```diff
@@ -109,6 +110,14 @@ defmodule Glimesh.Accounts.User do
     |> delete_change(:password)
   end
 
+  defp validate_displayname(changeset) do
+    username = get_field(changeset, :username)
+
+    changeset
+    |> validate_length(:displayname, min: 3, max: 50)
```
haydenmc, line 117: The complexity requirements for displayname/password/username are good candidates to be pulled out into a config file or constant somewhere. Might be outside the scope for this change though. We can open an issue on it if we think it's worth addressing later.

priv/repo/migrations/20200724162310_add_displayname_to_user_auth_table.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Glimesh.Repo.Migrations.AddDisplaynameToUserAuthTable do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :displayname, :string, default: nil
+    end
+    execute "UPDATE users SET displayname = username"
+    execute "UPDATE users SET username = lower(username)"
```
haydenmc, line 9: Why do we require usernames be lowercase? I personally like having capital letters in my URI - we can achieve case insensitive URIs without requiring all usernames be all lowercase.

haydenmc: If I'm in the minority here then that's fine, but I think it's worth discussing 😁

mja00: This is why displaynames were created, for capital letters. Read the issue this is tied to.

lib/glimesh/accounts/user.ex
mja00: This should be it's own PR/issue and not a part of this one.

haydenmc:
I've done some thinking after our convo yesterday, and I have some thoughts to share about #25.

I think it's worth considering an approach to #25 that utilizes case insensitive queries when resolving usernames rather than requiring the `displayname` field to serve as a case sensitive variant of `username`.

This can be achieved without compromising performance on Postgres by using a [non-deterministic collation](https://www.postgresql.org/docs/current/collation.html#COLLATION-NONDETERMINISTIC). Sadly, it doesn't look like this approach is agnostic to the DB provider and would be specific to Postgres likely through a custom `query` call in DB migration to apply the case-insensitive collation to the username field. This is definitely a compromise to consider.

However, I think there are some worthwhile benefits:
1. `displayname` doesn't necessarily have to match `username`, and can allow for the use of characters that may not be friendly to URI strings (whitespace, emoji, other unicode characters). This gives users more creative freedom with their name across the platform.
2. Removes the requirement of keeping the `username` and `displayname` fields in sync, reducing complexity of our DB model (it may not be very complex now, but having database constraints only by convention means we have to continue enforce this rule ourselves)

If we decide that the Postgres-specific solution to case insensitivity is too costly, and we'd like to continue with the separate field approach, I think we should consider making the following changes:
 - A. Rename the new field from `displayname` to `canonicalname` or something similar, so we can still benefit from point 1 above if we'd like to allow users to further customize their display names in the future
 - B. Enforce the requirement that `canonicalname` is the case-insensitive equivalent to `username` via a [database constraint](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#constraint/3) to address point 2 above.

Thank you for considering! :)

mja00:
After a solid discussion in Discord a plan to modify how this system will be implemented has been determined. The plan is as follows.

Continue to use a separate `username` and `displayname` field in the DB. Reversing the change of making `username` strictly lowercase. Solutions to the sensitivity: 
1) queries for username done case insensitively using collation 
2) inserting a new column for a lowercase version of the username and using a constraint to make sure they stay in sync

Neither option has been determined as the way to go and I'm looking forward to any discussion for either way of doing it.

haydenmc:
Option #2 has the pro of being DB provider agnostic, since Ecto doesn't have seem to have a mechanism for applying collations to fields (and even if it did, that approach is somewhat Postgres-specific).

So I lean a bit toward option #2 even though it 'pollutes' the model with an extra field.

mja00:
> So I lean a bit toward option 2 even though it 'pollutes' the model with an extra field.

Fully agree with this. While it doesn't sound good on paper it most likely is our best way of doing it. Adding another field shouldn't cause any confusion as long as it is named correctly.

lib/glimesh_web/templates/layout/_navbar.html.leex
```diff
@@ -31,7 +31,11 @@
           <li class="nav-item dropdown user-profile-dropdown">
               <a href="javascript:void(0);" class="nav-link dropdown-toggle user" id="userProfileDropdown"
                  data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
-                  <%= @current_user.username %>
+                  <%= if @current_user.displayname do %>
```
clone1018, line 34: Do we need this if statement if the user always has a displayname?

lib/glimesh_web/templates/user_settings/edit.html.eex
```diff
@@ -1,5 +1,5 @@
 <div class="container">
-    <h2>Your Profile</h2>
+    <h2 style="margin-top: 8px;">Your Profile</h2>
```
clone1018, line 2: `.mt-4` will give you some top margin

lib/glimesh_web/live/user_live/stream.ex
```diff
@@ -6,7 +6,7 @@ defmodule GlimeshWeb.UserLive.Stream do
   alias Glimesh.Presence
 
   def mount(%{"username" => streamer_username}, session, socket) do
-    case Streams.get_by_username(streamer_username) do
+    case Streams.get_by_username(String.downcase(streamer_username)) do
```
clone1018, line 9: Do we need the downcase here?

lib/glimesh_web/templates/layout/_navbar.html.leex
mja00: I can't remember for the life of me why I did that but definitely not needed.

lib/glimesh_web/live/user_live/stream.ex
mja00: Nope, will fix in next commit. Couldn't remember where I put those.

PR #34, opened 2020-07-25
Added development docker compose

omn0mn0m:


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "assets/css"]
 	path = assets/css
-	url = git@github.com:Glimesh/glimesh.tv-styles.git
+	url = ../../Glimesh/glimesh.tv-styles.git
```
clone1018, line 3: Is this right?

omn0mn0m: I would try it, but it worked for me in WSL 2 and in Manjaro.

omn0mn0m: Confirmed by cloning a new repository and trying to spin up. Also found a StackOverflow question confirming my solution: https://stackoverflow.com/questions/40841882/automatically-access-git-submodules-via-ssh-or-https

clone1018: Oh interesting, so it back two directories on GitHub... Not on your local filesystem.

Dockerfile.dev
```diff
@@ -0,0 +1,17 @@
...
+RUN mix local.hex --force
+RUN mix local.rebar --force
+
+# install node
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt-get install -y nodejs
+
+# prepare build dir
+WORKDIR /app
+EXPOSE 4000
```
haydenmc, line 17: Also `4001` for HTTPS?

omn0mn0m: Done :)

run.sh
```diff
@@ -0,0 +1,31 @@
...
+set -e
+
+# install dependencies
+mix deps.get
+
+# create and migrate database
+mix ecto.setup
+
+# install nodejs stuff
+cd assets && npm install
```
clone1018, line 12:
```suggestion
cd assets && npm ci
```

Can you reset the package lock, and change this script to use ci? That we we don't accidentally upgrade the whole repo?

PR #39, opened 2020-07-26
Mentions by username or @username

Hahkeye:
soon to have auto complete


lib/glimesh_web/live/chat_live/index.html.leex
```diff
@@ -10,16 +12,28 @@
                             <span>Welcome to chat! Follow the rules.</span>
                         </div>
                           <%= for chat_message <- @chat_messages do %>
-                            <div id="<%= chat_message.id %>" class="bubble you">
-                                <%= if @can_moderate and Glimesh.Chat.can_moderate?(@streamer, chat_message.user) === false do %>
-                                <i class="fas fa-times-circle" phx-click="timeout_user" phx-value-user="<%= chat_message.user.username %>"></i>
+                            <%=cond do %>
```
clone1018, line 15: Do you think we can shorten this to a CSS class that is added on mention? That way we don't have to duplicate the code above.

lib/glimesh/chat.ex
```diff
@@ -193,6 +193,9 @@ defmodule Glimesh.Chat do
     end
   end
 
+  def user_in_message(chat_message, user) do
```
clone1018, line 196:
```suggestion
  def user_in_message(username, chat_message) do
```
Might add some clarity to this function to flip the ordering to match the function name, and expand user to username.

lib/glimesh_web/live/chat_live/index.html.leex
Hahkeye: slightly confused but im pretty sure I made the change.

lib/glimesh/chat.ex
Hahkeye: fixed

lib/glimesh_web/live/chat_live/index.html.leex
```diff
@@ -2,25 +2,23 @@
 <div class="chat-system">
 
       <div class="chat-box">
+            <a target="_blank" href="https://donate.tiltify.com/@glimesh/glimesh-and-ikaros-launch-stream" class="btn btn-lg btn-trevor btn-block mb-2">Donate to The Trevor Project!</a>
```
clone1018, line 5: Just clean this up and we'll be good!

Hahkeye: Boom done

PR #54, opened 2020-08-02
Hcaptcha

Heiwa1580:
I think i added hcaptcha....


clone1018:
We can come back to this later!

PR #77, opened 2020-08-09
Profile Column Fix

MindlessPuppetz:
Fixing column alignment


clone1018:
Hey! Thank you for the PR, my apologies for the delay. I pulled this down and it aligns the logo properly, but it kinda makes the right footer content strange. 

image


Maybe if we adjust the columns inside that to remove the offset? What do you think?

MindlessPuppetz:
https://github.com/Glimesh/glimesh.tv/pull/96

This should fix it

PR #79, opened 2020-08-09
Basic 2FA authentication

### LINKED ISSUE #59: Add support for basic 2FA

wolfcomp:
Fixes #59

Basic 2FA authentication following the otpauth://totp/ standard

Feel free to roast my Elixir code and what not as this is my first time messing with it 
derp


lib/glimesh_web/templates/user_settings/edit.html.eex
```diff
@@ -196,4 +196,95 @@
...
+                    </div>
+                    <% end %>
+                </div>
+            </div>
+        </div>
+    </div>
+
 </div>
+
+<script>
```
clone1018, line 266: What does this code really power? Couldn't we handle this all on the server side to not have this javascript exist?

lib/glimesh_web/controllers/user_settings_controller.ex
```diff
@@ -77,6 +78,49 @@ defmodule GlimeshWeb.UserSettingsController do
...
+        {:ok, user} ->
+          conn
+          |> put_flash(:info, "2FA updated successfully.")
+          |> put_session(:user_return_to, Routes.user_settings_path(conn, :edit))
+          |> UserAuth.log_in_user(user)
+
+        {:error, changeset} ->
+          render(conn, "edit.html", tfa_changeset: changeset)
+      end
+    else
```
clone1018, line 99: These two case statements look similar, the only difference I see is the `tfa_token` param. You can do functional pattern matching, or since `user.tfa_token` is already nil pass it in directly

lib/glimesh_web/controllers/user_session_controller.ex
```diff
@@ -12,7 +13,17 @@ defmodule GlimeshWeb.UserSessionController do
...
+        %{"tfa" => tfa} = user_params
+
+        if Tfa.validate_pin(tfa, user.tfa_token) do
+          UserAuth.log_in_user(conn, user, user_params)
+        else
+          render(conn, "new.html", error_message: "Invalid 2FA code")
+        end
+      else
+        UserAuth.log_in_user(conn, user, user_params)
+      end
```
clone1018, line 26: This is a lot of nesting, how can we flatten it a bit?

lib/glimesh/accounts.ex
```diff
@@ -315,6 +315,57 @@ defmodule Glimesh.Accounts do
...
+  def update_tfa_temp(user, attrs) do
+    user
+    |> User.tfa_changeset(attrs)
+    |> Repo.update()
+  end
+
+  def update_tfa(user, pin, attrs) do
+    secret = user.tfa_token
+    IO.puts("attrs #{attrs.tfa_token}")
+    IO.puts("user #{user.tfa_token}")
```
clone1018, line 331: Remove for prod merge

clone1018:
This will also require a rebase against master. If you could wrap all the user facing strings in `dgettext(domain, string)` and then rerun `mix gettext.extract --merge` to generate the needed translation files, we'll be squared up!

lib/glimesh_web/templates/user_settings/edit.html.eex
wolfcomp: the reason I did this javascript code is that I couldn't find a method to safely store the 2FA hash code between requests and if there is a propper way to do this I can change it to use that method instead. but this is how the system know later what hash to use to check the 2FA codes with.

lib/glimesh/accounts.ex
wolfcomp: whops seems I forgot some debugging indeed XD

lib/glimesh_web/controllers/user_settings_controller.ex
wolfcomp: what I can do is switch how the if statement is as its only for removing the hash if the user already has a 2fa setup when they re enter the pin

lib/glimesh_web/controllers/user_session_controller.ex
wolfcomp: What I can do to change this is change where I figure if you have a token and pass or fail so instead of
```elixir
if user.tfa_token do
  %{"tfa" => tfa} = user_params

  if Tfa.validate_pin(tfa, user.tfa_token) do
    UserAuth.log_in_user(conn, user, user_params)
  else
    render(conn, "new.html", error_message: "Invalid 2FA code")
  end
else
  UserAuth.log_in_user(conn, user, user_params)
end
```
it could be 
```elixir
if Tfa.validate_pin(tfa, user.tfa_token) do
  UserAuth.log_in_user(conn, user, user_params)
else
  render(conn, "new.html", error_message: "Invalid 2FA code")
end```
where the check is inside the validate_pin

File: firstSetup.sh [hunk 5]
```diff
@@ -0,0 +1,13 @@
...
+sudo apt-get install -y elixir &&
+sudo apt-get install -y make &&
+sudo apt-get install -y build-essential &&
+sudo apt-get install -y nodejs &&
+sudo apt-get install -y npm &&
+sudo apt-get install -y inotify-tools &&
+sudo npm cache clean -f &&
+sudo npm install -g n &&
+sudo n stable &&
+PATH="$PATH"
```
clone1018, line 13: This was added but not documented, is this just for you?

wolfcomp: oh that got pushed. well this is if you have a complete new instance of ubuntu and want to start developing for it.

clone1018:
Checked it out locally and tested it, works very well!

I think ideally we split out the 2fa page into a net-new page, or convert this to a live view page so we can add fancy magic to show the 2fa if it's required for the user! However I'll probably merge this in now and we can make the change in a future PR.

clone1018:
Logged that here: https://github.com/Glimesh/glimesh.tv/issues/91

PR #87, opened 2020-08-12
Custom titles

### LINKED ISSUE #18: Allow user to customize their stream title
This should also be inside the live view, so viewers get updates as it changes.

Fields:
* Title 
* Description

mja00:
Fixes #18 

It done, it's pretty clean ngl. Video if it in action is on the dev channel in Discord.

### DIFF
--- a/lib/glimesh/streams.ex
+++ b/lib/glimesh/streams.ex
@@ -134,6 +134,7 @@ defmodule Glimesh.Streams do
     attrs = %{
       stream_title: title
     }
+
     streamer_meta = get_metadata_from_streamer(streamer)
 
     {:ok, results} =
@@ -153,6 +154,7 @@ defmodule Glimesh.Streams do
     attrs = %{
       stream_title: "My first stream!"
     }
+
     results =
       %StreamMetadata{
         streamer: streamer
@@ -163,7 +165,8 @@ defmodule Glimesh.Streams do
 
   def get_metadata_from_streamer(streamer) do
     data = Repo.get_by(StreamMetadata, streamer_id: streamer.id)
-    if !data do
+
+    unless data do
       create_metadata(streamer)
       get_metadata_from_streamer(streamer)
     else
@@ -178,7 +181,12 @@ defmodule Glimesh.Streams do
   defp broadcast_metadata({:error, _reason} = error, _event), do: error
 
   defp broadcast_metadata({:ok, data}, event) do
-    Phoenix.PubSub.broadcast(Glimesh.PubSub, "streams:#{data.streamer.id}:metadata", {event, data})
+    Phoenix.PubSub.broadcast(
+      Glimesh.PubSub,
+      "streams:#{data.streamer.id}:metadata",
+      {event, data}
+    )
+
     {:ok, data}
   end
 end
--- a/lib/glimesh/streams.ex
+++ b/lib/glimesh/streams.ex
@@ -134,13 +134,13 @@ defmodule Glimesh.Streams do
     attrs = %{
       stream_title: title
     }
-    streamerMeta = get_metadata_from_streamer(streamer)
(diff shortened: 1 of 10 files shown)


lib/glimesh_web/live/user_live/components/streamer_title.ex
```diff
@@ -0,0 +1,90 @@
...
+              <%= submit dgettext("streams", "Update Title"), class: "btn btn-primary mt-1" %>
+              <i class="far fa-edit" phx-click="toggle-edit" style="color: red"></i>
+            </div>
+          </h5>
+        <% end %>
+      <% end %>
+    <% else %>
+      <h5 class=""><span class="badge badge-danger">Live!</span> <%= @title %></h5>
+    <% end %>
+    """
```
clone1018, line 30: Hot DAMN! Look at this live view!

lib/glimesh_web/live/user_live/components/streamer_title.ex
```diff
@@ -0,0 +1,90 @@
...
+
+  alias Glimesh.Presence
+  alias Glimesh.Streams.StreamMetadata
+
+  @impl true
+  def render(assigns) do
+    ~L"""
+    <%= if @user do %>
+      <%= unless @user.username == @streamer.username do %>
+        <h5 class=""><span class="badge badge-danger">Live!</span> <%= @title %></h5>
```
clone1018, line 12: I see this line twice, can we compact it or add a new flag to replace these two if statements?

lib/glimesh_web/live/user_live/components/streamer_title.ex
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  @impl true
+  def handle_event("save", value, socket) do
+    case Glimesh.Streams.change_title(socket.assigns.streamer, value["stream_metadata"]["stream_title"]) do
+      {:ok, changetitle} ->
+        {:noreply,
+        socket
+        |> assign(:editing, false)
+        |> assign(:title, value["stream_metadata"]["stream_title"])
```
clone1018, line 74: You can pattern match this in the value if you feel like being more elixiry, then the function will never get called if someone is doing something weird. 

`handle_event("save", %{"stream_metadata" => %{"stream_title" => new_title_variable} }, socket)`

lib/glimesh/streams.ex
```diff
@@ -108,4 +108,58 @@ defmodule Glimesh.Streams do
   def count_following(user) do
     Repo.one!(from f in Followers, select: count(f.id), where: f.user_id == ^user.id)
   end
+
+  alias Glimesh.Streams.StreamMetadata
+
+  def change_title(streamer, title) do
```
clone1018, line 114: `change_` is for generating a unsaved changeset like you have here: https://github.com/Glimesh/glimesh.tv/pull/87/files#diff-677844d838361a57c68f062b05853dd6R48

`update_` is for actually committing the transaction. Checkout accounts.ex `change_user_profile` vs `update_user_profile`

lib/glimesh/streams.ex
```diff
@@ -108,4 +108,58 @@ defmodule Glimesh.Streams do
...
+    if !data do
+      create_metadata(streamer)
+      get_metadata_from_streamer(streamer)
+    else
+      data
+    end
+  end
+
+  def subscribe_metadata do
+    Phoenix.PubSub.subscribe(Glimesh.PubSub, "streams")
```
clone1018, line 156: If we namespace `streams:user_id` we don't have to do any streamer checks after this. Not sure the implications of so many channels though?

clone1018: We can wait and do this refactor later though, since chat / view count suffer from it right now too. We want to standardize this for the API so I can list to the right channels...

clone1018:
Awesome stuff! Thanks again

PR #88, opened 2020-08-12
Updated the internationalization and imported Norwegian

mja00:
Fixed some missing gettexts and issues with LiveView
Added Norwegian as a supported language with full translation support(Thanks Datagutt!)


mja00:
> full translation support

To be fair it was full translation till I added new terms. Those have been also added to POEditor!

PR #96, opened 2020-08-14
fix(profile): fix footer spacing on profile page

MindlessPuppetz:
Change offset to 2 instead of 3 to fix footer info

### DIFF
--- a/lib/glimesh_web/templates/layout/root.html.leex
+++ b/lib/glimesh_web/templates/layout/root.html.leex
@@ -62,7 +62,7 @@
         </div>
         <div class="col-md-9">
           <div class="row">
-            <div class="col-6 col-md-3 offset-md-3">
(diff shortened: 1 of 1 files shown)


lib/glimesh_web/templates/layout/root.html.leex
```diff
@@ -62,7 +62,7 @@
         </div>
         <div class="col-md-9">
           <div class="row">
-            <div class="col-6 col-md-3 offset-md-3">
+            <div class="col-6 col-md-3 offset-md-2">
               <h5>Social</h5>
```
mja00, line 66: To resolve the current conflict you'll just have to do:
`<h5><%= dgettext("footer", "Social") %> </h5>`

clone1018:
Fixed in https://github.com/Glimesh/glimesh.tv/pull/165 ! Thanks for your support!

PR #124, opened 2020-08-26
Add specific page for 2FA auth once a user logs in

clone1018:


mja00:
I know I beat it into your head but now it's just gettext(string) instead of dgettext(domain, string). That way importing translations is easier and faster for me.

clone1018:
This code is from before we moved to gettext, I just rebased it so far. I'll get it up to date with our current standards before it moves out of draft :)

clone1018:
Oops, forgot about that. Pushed!

PR #125, opened 2020-08-26
Added 2000+ badwords to filter out.

Zaeo:
The original list is commented out.


config/badwords.exs
```diff
@@ -50,387 +50,236 @@ config :glimesh,
...
+    "cock-head", "cock-sucker", "cockhead", "cocks", "cocksucker", "crap", "cum", "cunt", "cunts", "cunty", "cuntz", "d4mn", "dab",
+    "damn", "daygo", "dego", "dick", "dicksucker", "dicklicker", "dickfucker", "dike", "dild0", "dild0s", "dildo", "dildos", "dilld0",
+    "dilld0s", "dirsa", "dominatricks", "dominatrics", "dominatrix", "dupa", "dyke", "dziwka", "ejackulate", "ejakulate", "Ekrem",
+    "ekto", "enculer", "enema", "f u c k", "f u c k e r", "faen", "fag", "fag1t", "faget", "fagg1t", "faggit", "faggot", "fagit", "fags", #Hate Speech Present
+    "fagz", "faig", "faigs", "fanculo", "fanny", "fart", "fatass", "fcuk", "feces" "feg", "Felcher", "ficken", "fiddler", "fitt", 
+    "flikker", "flipping the bird", "fokken", "foreskin", "fotze", "fu\(", "fuck", "fucker", "fuckin", "fucking", "fucks","fudge packer",
+    "fuk", "Fukah", "fuken", "fuker", "fukin", "fukk", "fukkah", "fukken", "fukker", "fukkin", "futkretzn", "fux0r", "g00k", 
+    "gay","gayboy","gayboi","gaygirl", "gaygurl", "gays", "gayz", "god-damned", "gook", "guiena", "h00r", "h0ar", "h0r", "h0re", "h4x0r", #Hate Speech Present
```
haydenmc, line 112: I'm not sure `gay` and `lesbian` should be considered bad words. I understand they can be used in offensive contexts, but the words themselves are not bad.

wolfcomp: There are also a lot of other words in here that aren't bad in them self but also dependent on the context. Possible solution would be to split it up into age categories as some of them allows for more lenient usage of these words than others and based on that filter them out.

PR #128, opened 2020-08-28
:seedling: Add per-environment seeding as part of application setup

shrink:
The application can't currently launch through the docker-compose setup because `priv/repo/seeds.exs` doesn't exist. I have added a seed file for each environment, which can in turn include one or more other seed files. I have added the current seeds (categories and streams) through a wildcard inclusion for both dev and test but not production.


clone1018:
Thanks for the work!

Issue #112, opened 2020-08-20
Implement site wide account banning 


clone1018:
IP Address? Simple account ban? Who knows...

mja00:
Personally send Paco to the person's house and tell them to knock it off. :)

mja00:
Think I'll take this as my next issue. Implement simple account bans at first and then maybe expand upon it.

PR #132, opened 2020-09-07
Added site wide account bans. 

mja00:
Adds site wide account bans which was requested in issue #112 

Currently no way of issuing an account ban other than setting is_banned to true for a user in the DB. 

Once that is true the user will no longer be able to login and their current session will be logged out when they navigate to a new page.

Still have to add a check when displaying a user's profile, and they're still able to chat currently if marked as banned.

### DIFF
--- a/lib/glimesh/accounts.ex
+++ b/lib/glimesh/accounts.ex
@@ -88,6 +88,11 @@ defmodule Glimesh.Accounts do
   """
   def get_user!(id), do: Repo.get!(User, id)
 
+  def is_user_banned_by_username(username) do
+    user = Repo.get_by(User, username: username)
+    user.is_banned
+  end
+
   ## User registration
 
   @doc """
--- a/lib/glimesh/accounts.ex
+++ b/lib/glimesh/accounts.ex
@@ -46,12 +46,12 @@ defmodule Glimesh.Accounts do
       nil
 
   """
-  def get_by_username(username) when is_binary(username) do
-    Repo.get_by(User, username: username)
+  def get_by_username(username, ignore_banned \\ false) when is_binary(username) do
+    Repo.get_by(User, username: username, is_banned: ignore_banned)
   end
 
-  def get_by_username!(username) when is_binary(username) do
-    Repo.get_by!(User, username: username)
+  def get_by_username!(username, ignore_banned \\ false) when is_binary(username) do
+    Repo.get_by!(User, username: username, is_banned: ignore_banned)
   end
 
   @doc """
--- a/lib/glimesh/accounts.ex
+++ b/lib/glimesh/accounts.ex
@@ -88,7 +88,7 @@ defmodule Glimesh.Accounts do
   """
   def get_user!(id), do: Repo.get!(User, id)
 
-  def is_user_banned_by_username(username) do
+  def is_user_banned_by_username?(username) do
     user = Repo.get_by(User, username: username)
     user.is_banned
   end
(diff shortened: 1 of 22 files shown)


lib/glimesh_web/templates/user_session/new.html.eex
```diff
@@ -238,7 +238,7 @@
                     <%= form_for @conn, Routes.user_session_path(@conn, :create), [as: :user, class: "text-left"], fn f -> %>
                         <%= if @error_message do %>
                             <div class="alert alert-danger">
-                              <p><%= @error_message %></p>
+                              <center><%= @error_message %></center>
```
clone1018, line 241:
```suggestion
                              <div class="text-center"><%= @error_message %></div>
```

Center? Are you in 1992?

mja00: Thanks you fixed it for me

mja00:
Wanna add a reason field when being banned.

priv/repo/migrations/20200907160029_site-wide-account-bans.exs
```diff
@@ -0,0 +1,10 @@
+defmodule :"Elixir.Glimesh.Repo.Migrations.Site-wide-account-bans" do
```
clone1018, line 1: This looks funky, can rename?

clone1018 (changes requested):

priv/repo/migrations/20200907160029_site-wide-account-bans.exs
mja00: Good question. Lemme check.

mja00: The answer was yes

mja00:
Had to rebase due to this branch still being before our consensus on doing all translation updates in separate PRs, should be the last rebase needed.

lib/glimesh/accounts.ex
```diff
@@ -88,6 +88,11 @@ defmodule Glimesh.Accounts do
   """
   def get_user!(id), do: Repo.get!(User, id)
 
+  def is_user_banned_by_username(username) do
```
clone1018, line 91:
```suggestion
  def is_user_banned_by_username?(username) do
```
I really like the convention of using trailing question marks on boolean functions, what do you think?

lib/glimesh_web/controllers/user_session_controller.ex
```diff
@@ -11,12 +11,24 @@ defmodule GlimeshWeb.UserSessionController do
 
   def create(conn, %{"user" => %{"email" => email, "password" => password} = user_params}) do
     if user = Accounts.get_user_by_email_and_password(email, password) do
-      if user.tfa_token do
-        conn
-        |> put_session(:tfa_user_id, user.id)
-        |> render("tfa.html", error_message: nil)
-      else
-        UserAuth.log_in_user(conn, user, user_params)
+      case user.is_banned do
```
clone1018, line 14: Why case instead of if?

lib/glimesh_web/controllers/user_session_controller.ex
```diff
@@ -11,12 +11,24 @@ defmodule GlimeshWeb.UserSessionController do
...
+          end
+
+        true ->
+          render(conn, "new.html",
+            error_message:
+              gettext(
+                "User account is banned. Please contact support at %{email} for more information.",
+                email: "support@glimesh.tv"
+              )
+          )
```
clone1018, line 31: Looking at this function, it's kinda big, any ideas on splitting it out? If not we can keep moving and come back later.

clone1018 (changes requested):

lib/glimesh/accounts.ex
mja00: I think that would be a good convention, quick and easy way of seeing what the function is going to return.

clone1018 (approved):

PR #139, opened 2020-09-10
Fixed path issue for env seeds

omn0mn0m:
Tested and works in my work env from fresh install.


cryptixishere (commented):
These changes work for Docker Compose (dev-only) and WSL

PR #141, opened 2020-09-10
Add a Markdown toolbar to help with editing

omn0mn0m:
For those that don't already know Markdown. It is using GitHub's toolbar.

image


omn0mn0m:
Please accept #139 before merging.

clone1018:
Thanks for making this!

One question so far, does our markdown support checkboxes? Should we remove that option?

mja00 (approved):

omn0mn0m:
> Thanks for making this!
> 
> One question so far, does our markdown support checkboxes? Should we remove that option?

It does not :( Kinda lame, could see some benefit in people being able to check off milestones and such.

PR #129, opened 2020-08-29
Feature OAuth2

wolfcomp:
Its a long list of files added

PR #143, opened 2020-09-11
Feature: OAuth 2.0 & Developer Apps (#129 & #142 & #145 & #146)

clone1018:
* Add yarns lock file to ignore list and fix an ex file missing a html end sign

* The amount of time it took me for this was worth it

* user end control over applications implemented

* Implementation of RFC6749 also known as Oauth2 is done

* Fixing yelling from "Code Readability"

* Oauth Introspec done should be everything for oauth in terms of requesting a token now to actually auth one

* Fix credo not linking multi alias as alias sorting is not a "Code Readability" issue and can be solved while doing reviews

* Api now accepts Oauth2 Auth header as authorization

* All the language feeding.

* Missed one file

* Oops forgot this one

* Final fixes

* Missed some text-alligns to be of class text-center

* Forgot one spacer

* Fix some button styling

* one final spacer for the buttons

* Add streamkey scope

* Add migration

* Update for it to allow all scopes that exists once the oauth app is created will break if we add more later but should be enough for now *going to rewrite everything after chat rework*

* Adding 3rd Party Applications  (#142)

* Making a generic apps concept, adding some initial framework

* Functional tests

* Wrapping up formatting of applications

* Adding designed grants pages

* Adding translations

* Removing translations

Co-authored-by: Luke Strickland <luke@axxim.net>


clone1018:
image

PR #160, opened 2020-09-16
Added a followers page for profiles.

mja00:
Just need to click on the followers count to pull it up!

### DIFF
--- a/lib/glimesh_web/live/user_live/profile.html.leex
+++ b/lib/glimesh_web/live/user_live/profile.html.leex
@@ -50,12 +50,12 @@
 
                     <div class="simple--counter-container mt-4">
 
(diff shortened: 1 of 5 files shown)


lib/glimesh_web/live/user_live/profile.html.leex
```diff
@@ -50,12 +50,12 @@
 
                     <div class="simple--counter-container mt-4">
 
-                        <div class="counter-container">
+                        <a href="profile/followers"><div class="counter-container">
```
clone1018, line 53: Can this be a proper phoenix link so we don't suffer from accidental url munging?

mja00: Sure

PR #161, opened 2020-09-16
Fixing redirect bugs, profile pictures, and display name changes

clone1018:
1. Fix the profile picture being egg shaped
2. Fix 2fa redirect
3. link api documentation in “check out our API documentation”
4. Update open graph image with something not three months out of date
5. Fixed display name changes not always working with certain usernames
6. Adding black and white version of the 2fa barcode


lib/glimesh_web/templates/user_applications/index.html.eex
```diff
@@ -23,6 +23,9 @@
           <p>
             <%= gettext("Want to build a chat bot, a online tool, or other service that integrates with Glimesh? You're in the right place! Create an Application to get started and checkout our API documentation if you need some guidance.") %>
```
mja00, line 24:
```suggestion
            <%= gettext("Want to build a chat bot, a online tool, or other service that integrates with Glimesh? You're in the right place! Create an Application to get started and check out our API documentation if you need some guidance.") %>
```

mja00 (changes requested):
Just that one thing and it’s good.

lib/glimesh_web/templates/user_applications/index.html.eex
clone1018: @mja00 I was worried this would mess up the updates you just pushed out, we good?

PR #165, opened 2020-09-17
Fixed an unclosed anchor tag

omn0mn0m:
Footer should be better aligned now :)


omn0mn0m:
Please accept #139 before merging.

Issue #163, opened 2020-09-17
Application redirect URIs do not allow non-HTTPS connections


SaviorXTanren:
I'm unable to create an application for our bot because the allowed redirect URIs restrict usage of HTTP urls. Mix It Up is a desktop-only bot and there is no server authentication. Everything happens directly on the user's machine for the OAuth redirect.

I'm OK with the fact that it's not something allowed directly in the UI, but there should be an option to allow for non-HTTPS urls to be used such as filing a support ticket to have someone manually update it in the backend. Otherwise we will either need to stand  up an authentication server (costly and more security hurdles) or do some form of browser container manipulation to get the parameters from the url redirect.

![image](https://user-images.githubusercontent.com/752903/93535107-276a9e80-f8fb-11ea-9e6d-8da6f6dc8799.png)

SaviorXTanren:
Awesome, thanks @wolfcomp !

PR #168, opened 2020-09-19
Allow http links with redirect uris

wolfcomp:
Fixes: #163

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -45,7 +45,8 @@ config :ex_oauth2_provider, ExOauth2Provider,
   optional_scopes: ~w(email chat streamkey),
   authorization_code_expires_in: 600,
   access_token_expires_in: 21600,
-  grant_flows: ~w(authorization_code client_credentials implicit_grant)
+  grant_flows: ~w(authorization_code client_credentials implicit_grant),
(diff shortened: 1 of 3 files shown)


mja00 (approved):

haydenmc:
I think we should restrict this exemption to only localhost redirects for desktop/native apps. 

We shouldn't encourage web apps to exchange creds over plaintext. 

https://docs.microsoft.com/en-us/azure/active-directory/develop/reply-url#localhost-exceptions

haydenmc (changes requested):
I think we should restrict this exemption to only localhost redirects for desktop/native apps. 

We shouldn't encourage web apps to exchange creds over plaintext. 

Example:
https://docs.microsoft.com/en-us/azure/active-directory/develop/reply-url#localhost-exceptions

SaviorXTanren:
That's a fairly common practice I see and that's perfectly meets our requirements. As long as we're able to specify localhost and a specific port for it, that works. (EX: "http://localhost:8091")

clone1018:
Added logic to ensure that you're only using an approved localhost url.

haydenmc (approved):

PR #189, opened 2020-09-25
Adding start_stream and end_stream functionality

clone1018:


test/glimesh/streams_test.exs
```diff
@@ -203,6 +203,7 @@ defmodule Glimesh.StreamsTest do
         audio_codec: "mp3",
```
haydenmc, line 203: Might make sense to change audio codec and video codec to [supported](https://github.com/Glimesh/janus-ftl-plugin/blob/master/VideoCodec.h#L33) [values](https://github.com/Glimesh/janus-ftl-plugin/blob/master/AudioCodec.h#L33) - currently `opus` and `h264`.

haydenmc: I'd also planned on reporting bitrate in bps instead of kbps - might be worth clarifying the units, either in the member name or with a comment.

haydenmc: (sorry, should've brought that up in the previous change that first introduced these fields)

haydenmc (approved):

test/glimesh/streams_test.exs
clone1018: https://github.com/Glimesh/glimesh.tv/issues/195

PR #206, opened 2020-09-28
Adding ability to stop stream by stream_id

clone1018:


lib/glimesh/resolvers/streams_resolver.ex
```diff
@@ -38,13 +38,23 @@ defmodule Glimesh.Resolvers.StreamsResolver do
     {:error, "Access denied"}
   end
 
+  def end_stream(_parent, %{stream_id: stream_id}, %{context: %{is_admin: true}}) do
+    stream = Streams.get_stream!(stream_id)
+
+    Streams.end_stream(stream)
+  end
+
   def end_stream(_parent, %{channel_id: channel_id}, %{context: %{is_admin: true}}) do
```
haydenmc, line 47: I still think it's dangerous to allow ending of arbitrary stream instances by channel ID - if we allow this, we ought to plumb down some notification to anyone listening (FTL instances) that the stream IDs they're operating on have been invalidated so they don't just chug along happily not knowing that the service has decided that they're gone.

lib/glimesh/resolvers/streams_resolver.ex
```diff
@@ -38,13 +38,23 @@ defmodule Glimesh.Resolvers.StreamsResolver do
...
     {:ok, stream}
   end
 
+  def end_stream(_parent, _args, %{context: %{is_admin: true}}) do
+    {:error, "Must specify channelId or streamId"}
+  end
+
   def end_stream(_parent, _args, _resolution) do
     {:error, "Access denied"}
   end
```
haydenmc, line 60: Can we get this for `log_stream_metadata` as well?

haydenmc (commented):
Couple comments, looks good

lib/glimesh/resolvers/streams_resolver.ex
clone1018: Added stream_id as the only allowed input.

haydenmc (approved):

Issue #217, opened 2020-10-06
Add "My Channel" to the dropdown on the right 


mja00:

cvandermeer:
This will link to the users public profile?

I don't have access to push a branch.

PR #218, opened 2020-10-07
Add new options to the dropdown on the navbar

mja00:
Added a channel link and a profile link

Solves #217

### DIFF
--- a/lib/glimesh/streams.ex
+++ b/lib/glimesh/streams.ex
@@ -131,7 +131,7 @@ defmodule Glimesh.Streams do
         on: c.user_id == u.id,
         where: u.id == ^user.id,
         where: c.inaccessible == false
-    )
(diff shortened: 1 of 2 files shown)


lib/glimesh/streams.ex
```diff
@@ -125,7 +125,13 @@ defmodule Glimesh.Streams do
   end
 
   def get_channel_for_user!(user) do
-    Repo.get_by(Channel, user_id: user.id) |> Repo.preload([:category, :user])
```
clone1018, line 128: Did we need the  |> Repo.preload([:category, :user]) for something?

mja00: Good question, I'll add it back into the the function. Since it didn't break anything not being there but it could be an edge case that breaks from it.

PR #225, opened 2020-10-12
Making the GCT dashboard

mja00:
Added a dashboard/panel for our community team. 

Currently allowed them to:
- Edit users
- Edit profiles
- Edit channels
- Ban users

Also incorporates an audit log for any and all actions done on the dashboard.


lib/glimesh/community_team.ex
```diff
@@ -0,0 +1,124 @@
...
+
+  def get_global_access_level(level_int) do
+    if level_int, do: Enum.at(Integer.digits(level_int), 0, 0), else: 0
+  end
+
+  def get_billing_override(level_int) do
+    if level_int, do: Enum.at(Integer.digits(level_int), 1, 0), else: 0
+  end
+
+  def generate_update_user_profile_more_details(user, user_params) do
```
mja00, line 82: For these(since I know they are terribly ugly), the plan is to eventually generate these strings dynamically based on the changed data. Most likely will need to compare the two maps and then find the differences. This was just a quick way of getting the needed information without spending too long on it.

assets/css/glimesh/components/sidebar.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/* On screens that are less than 700px wide, make the sidebar into a topbar */
+@media screen and (max-width: 700px) {
+  .sidebar {
+    width: 100%;
+    height: auto;
+    position: relative;
+  }
+  .sidebar a {float: left;}
```
haydenmc, line 40: Consider using `flexbox` for laying out elements to one side or another, `float` is largely considered obsolete for this, and is really only meant for things like floating text around an image in an article.
https://stackoverflow.com/a/9777248

lib/glimesh/accounts/user.ex
```diff
@@ -203,6 +208,31 @@ defmodule Glimesh.Accounts.User do
     |> cast_attachments(attrs, [:avatar])
   end
 
+  def big_scary_changeset(user, attrs) do
```
haydenmc, line 211: This is not a very descriptive name

mja00: It used to be more descriptive. Clone wanted it named in a way that made it so it couldn't be used accidentally. Used to be called gct_user_changeset iirc.

mja00: Looks to be user_changeset before swapping. https://github.com/Glimesh/glimesh.tv/pull/225/commits/bf012c90c501ca8c27ab4270e28964ab41493ff6

haydenmc: Can we name it in a way where it's clear what the intended use is? The other names clearly indicate their use - `stripe_changeset`, `profile_changeset`, `password_changeset`, etc.

haydenmc: `gct_user_changeset` seems reasonable to me but cc @clone1018

lib/glimesh/accounts/user.ex
```diff
@@ -290,4 +320,19 @@ defmodule Glimesh.Accounts.User do
       add_error(changeset, :tfa, "Invalid 2FA code")
     end
   end
+
+  @doc """
+  Validates the GCT access level
+  """
+  def validate_gct_level(changeset, needed_amount) do
```
clone1018, line 327: What is this exactly? Looks like there's a couple of warnings as well.

lib/glimesh/community_team.ex
```diff
@@ -0,0 +1,124 @@
...
+      4 -> "Manager"
+      3 -> "Team Lead"
+      2 -> "Team Member"
+      1 -> "Trial Member"
+      _ -> "None"
+    end
+  end
+
+  def can_edit_user(user) do
+    if get_global_access_level(user.gct_level) >= 3, do: true, else: false
```
clone1018, line 21: Confused about these functions as well, shouldn't we make the permissions individual knobs? `can_edit_user`, `can_ban_user`, etc?

lib/glimesh/community_team.ex
```diff
@@ -0,0 +1,124 @@
...
+  def create_audit_entry(user, attrs \\ %{action: "None", target: "None"}) do
+    %AuditLog{
+      user: user
+    }
+    |> AuditLog.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  def list_all_audit_entries(include_verbose? \\ false, params \\ []) do
+    _entries =
```
clone1018, line 53: _entires should be removed.

lib/glimesh/community_team/audit_log.ex
```diff
@@ -0,0 +1,27 @@
...
+  @moduledoc false
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  schema "gct_audit_log" do
+    belongs_to :user, Glimesh.Accounts.User
+
+    field :action, :string
+    field :target, :string
+    field :verbose_required?, :boolean
```
clone1018, line 11: Does this actually put a question mark in the SQL column name?

lib/glimesh_web/templates/gct/edit_user_profile.html.eex
```diff
@@ -0,0 +1,151 @@
+<div class="container">
+    <div class="alert alert-danger mt-4 mb-1"><h2 class="mb-0 text-center"><%= gettext("WARNING YOU'RE CURRENTLY EDITING %{user}'s PROFILE SETTINGS", user: @user.username) %></h2></div>
+
```
clone1018, line 3: How can we not have this page duplicated since its the same view as the user profile?

lib/glimesh_web/templates/gct/lookup_user.html.eex
```diff
@@ -0,0 +1,166 @@
...
+                        <%= textarea :user, :ban_reason, [value: @user.ban_reason,class: "form-control mb-4", id: "textarea_edit", rows: 5, disabled: true]  %>
+                    </div>
+                </div>
+                <!-- Row end -->
+                <% end %>
+            </div>
+        </div>
+        </div>
+    </div>
+    <%= if @view_billing? do %>
```
clone1018, line 113: These don't seem to be in sync with what we show the user. Had some issues showing payout history, would be better to reuse the existing view for this.

clone1018 (commented):
First round done, mostly architecture questions.

lib/glimesh/accounts/user.ex
mja00: This makes sure the gct access level is the correct amount of digits long. I changed somewhat how the access levels works and each digit corresponds to a different permission. Right now the 2 digits are: global GCT access and billing access. With the billing access not implemented yet.

lib/glimesh/community_team.ex
mja00: That is also possibly but I'm more afraid of the clutter that would cause. Having it set to a number level allows for more permissions to be added to someone without having to go through and edit the entire team.

lib/glimesh/community_team/audit_log.ex
mja00: image
Yes :)

lib/glimesh/community_team.ex
mja00: Yeah I think at one point I was using the variable and just forgot to completely remove it.

lib/glimesh_web/templates/gct/edit_user_profile.html.eex
mja00: Was thinking of turning the card on that page into a component that is shared by both the user profile and the GCT side.

lib/glimesh_web/templates/gct/lookup_user.html.eex
mja00: Could use same as above.

lib/glimesh/accounts.ex
```diff
@@ -306,6 +312,10 @@ defmodule Glimesh.Accounts do
     User.profile_changeset(user, attrs)
   end
 
+  def change_user(user, attrs \\ %{}) do
```
clone1018, line 315: The naming here isn't what I would expect for the changeset it's calling.

lib/glimesh/accounts.ex
```diff
@@ -324,6 +334,12 @@ defmodule Glimesh.Accounts do
     |> Repo.update()
   end
 
+  def update_user(user, attrs) do
```
clone1018, line 337: Same as comment above, this is in the account's context and is called update_user but it uses the GCT changeset?

lib/glimesh/accounts.ex
```diff
@@ -534,5 +550,40 @@ defmodule Glimesh.Accounts do
 
   def can_stream?(user), do: user.can_stream
 
-  def can_use_payments?(user), do: user.can_payments
+  def can_use_payments?(user) do
```
clone1018, line 553: Why the switch away from the short do: ?

lib/glimesh/accounts.ex
```diff
@@ -534,5 +550,40 @@ defmodule Glimesh.Accounts do
...
+    end
+  end
+
+  def unban_user(user) do
+    user
+    |> User.gct_user_changeset(%{is_banned: false, ban_reason: nil})
+    |> Repo.update()
+  end
+
+  def throw_error_on_action(error_message, attrs, action) do
```
clone1018, line 577:
```suggestion
  defp throw_error_on_action(error_message, attrs, action) do
```

lib/glimesh_web/live/gct_live/components/button_array.ex
```diff
@@ -0,0 +1,132 @@
...
+
+  @impl true
+  def mount(_params, %{"admin" => admin, "user" => user}, socket) do
+    can_ban =
+      case Bodyguard.permit(Glimesh.CommunityTeam, :can_ban, admin, user),
+        do:
+          (
+            :ok -> true
+            {:error, :unauthorized} -> false
+          )
```
clone1018, line 58: Can be shortened to: 
```suggestion
    can_ban = Bodyguard.permit?(Glimesh.CommunityTeam, :can_ban, admin, user)
```

same with below can_'s

lib/glimesh_web/live/gct_live/components/button_array.ex
```diff
@@ -0,0 +1,132 @@
...
+     |> assign(:user, user)
+     |> assign(:show_ban, false)
+     |> assign(:can_ban, can_ban)
+     |> assign(:can_edit_user, can_edit_user)
+     |> assign(:can_edit_profile, can_edit_user_profile)}
+  end
+
+  @impl true
+  def handle_event("ban", %{"ban_reason" => ban_reason}, socket) do
+    {:ok, _} = Accounts.ban_user(socket.assigns.user, ban_reason)
```
clone1018, line 88: This should be behind a permissions check.

lib/glimesh/accounts.ex
```diff
@@ -534,5 +550,40 @@ defmodule Glimesh.Accounts do
 
   def can_stream?(user), do: user.can_stream
 
-  def can_use_payments?(user), do: user.can_payments
+  def can_use_payments?(user) do
+    user.can_payments
+  end
+
+  def ban_user(user, reason) do
+    case reason do
```
clone1018, line 558: These unsafe functions should be locked behind a permission check here instead of in the controller. This helps ensure we're always writing safe code. An example can be found here: https://github.com/Glimesh/glimesh.tv/blob/dev/lib/glimesh/stream_moderation.ex#L122

lib/glimesh/community_team.ex
clone1018: Still need to be removed.

lib/glimesh_web/controllers/gct_controller.ex
```diff
@@ -0,0 +1,310 @@
...
+             conn.assigns.current_user,
+             params["query"]
+           ) do
+      unless params["query"] == "",
+        do:
+          CommunityTeam.create_audit_entry(conn.assigns.current_user, %{
+            action: "lookup",
+            target: params["query"],
+            verbose_required: true
+          })
```
clone1018, line 51: The above code is very hard to decipher, how can we make it clearer?

lib/glimesh_web/live/gct_live/components/button_array.ex
clone1018: Same for unban_user below.

clone1018 (changes requested):
Feels like we're almost there! Couple more improvements we can make.

lib/glimesh_web/templates/gct/lookup_user.html.eex
mja00: I've temporarily disabled the payments/payouts history till I get the existing views working(most likely the next PR after this is made).

mja00:
Ready for round 3(maybe 4?)

clone1018 (approved):

clone1018:
Approved, PULL THE LEVER!

PR #233, opened 2020-10-18
Added confetti to clone's stream if it is his birthday

mja00:
:) This also introduces a library that can be used for confetti in the future for subs and resubs if wanted. Can be installed and used through NPM but wanted the faster/easier way.


mja00:
This funny but needs to be implemented properly. :)

PR #250, opened 2020-11-09
Update homepage_live.html.leex

Heiwa1580:
Changed some wording around

### DIFF
--- a/lib/glimesh_web/live/homepage_live.html.leex
+++ b/lib/glimesh_web/live/homepage_live.html.leex
@@ -61,7 +61,7 @@
             <img src="/images/glimesh-verified.png" alt="" class="" style="max-height: 250px">
           </div>
           <div class="col-md-8 order-md-1">
-            <h2 class="featurette-heading"><%= gettext("Subscriptions for Everyone.") %> <br><small class="text-muted"><%= gettext("Level the playing field.") %></small></h2>
+            <h2 class="featurette-heading"><%= gettext("Level the playing field.") %> <br><small class="text-muted"><%= gettext("Subscriptions aren't just for the pros!") %></small></h2>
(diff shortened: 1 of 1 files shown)


lib/glimesh_web/live/homepage_live.html.leex
```diff
@@ -61,7 +61,7 @@
             <img src="/images/glimesh-verified.png" alt="" class="" style="max-height: 250px">
           </div>
           <div class="col-md-8 order-md-1">
-            <h2 class="featurette-heading"><%= gettext("Subscriptions for Everyone.") %> <br><small class="text-muted"><%= gettext("Level the playing field.") %></small></h2>
+            <h2 class="featurette-heading"><%= gettext("Level the playing field.") %> <br><small class="text-muted"><%= gettext("Subscriptions aren't just for the pros!") %></small></h2>
```
clone1018, line 64: Thought about this a lot, I don't like "Subscriptions aren't just for the pros!". It kinda insinuates that streamers without subscriptions are not professional on other platforms. 

What else can we choose for this sentence?

clone1018 (changes requested):

lib/glimesh_web/live/homepage_live.html.leex
ShinraCorpMixer: I have to agree, another way of structuring and saying it would be: "Level the playing field, become your own professional streamer!" Or something like that, sounds corny but that's what I got for now.

Heiwa1580: Yeah I agree on that one. What about just saying

We're all pros heres.

clone1018:
@Heiwa1580 Still interested in this change?

Heiwa1580:
I personally don't mind the old wording. I mostly made this change because others seemed to want a different idea thrown out there. Do with it what you think is best I personally have no issue with the original wording.

clone1018:
Closing due to inactivity.

PR #252, opened 2020-11-11
Adding credits page to list our subscribers & open source projects

clone1018:
image

### DIFF
--- a/lib/glimesh/credits.ex
+++ b/lib/glimesh/credits.ex
@@ -32,7 +32,7 @@ defmodule Glimesh.Credits do
       {"libssl",
        "OpenSSL is a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols."},
       {"libavcodec",
-       "Generic encoding/decoding framework and contains multiple decoders and encoders for audio, video and subtitle streams, and several bitstream filters."},
+       "Library from ffmpeg providing a encoding/decoding framework and multiple decoders and encoders for audio, video and subtitle streams, and several bitstream filters."},
       {"spdlog", "Very fast, header-only/compiled, C++ logging library."},
       {"cpp-httplib", "A C++11 single-file header-only cross platform HTTP/HTTPS library."},
       {"Catch2",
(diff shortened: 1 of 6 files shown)


clone1018:
@haydenmc can you check out the deps for the FTL projects? Stole it from the masons.

lib/glimesh_web/templates/about/credits.html.eex
```diff
@@ -0,0 +1,103 @@
...
+                style="width: 80% !important;" class="img-fluid mb-2 img-avatar">
+        </div>
+        <% end %>
+    </div>
+
+    <% end %>
+</div>
+
+<h2><%= gettext("Open Source Libraries") %></h2>
+<p><%= gettext("We use numerous open source projects to power Glimesh, and without them this project would never have been possible. This is not a comprehensive list unfortunately. You can find a full list by browsing our core repositories:") %>
```
mja00, line 41:
```suggestion
<p><%= gettext("We use numerous open source projects to power Glimesh, and without them this project would never have been possible. This is not a comprehensive list, unfortunately. You can find a full list by browsing our core repositories:") %>
```

Just a small grammatical thing :)

mja00 (commented):
Looking good :)

haydenmc (commented):
libcrypto is considered part of openssl, it might make sense to only mention openssl here

lib/glimesh/credits.ex
```diff
@@ -0,0 +1,54 @@
...
+      {"glib-2.0",
+       "GLib provides the core application building blocks for libraries and applications written in C."},
+      {"libsrtp2",
+       "Implementation of the Secure Real-time Transport Protocol (SRTP), the Universal Security Transform (UST), and a supporting cryptographic kernel. "},
+      {"jansson", "Jansson is a C library for encoding, decoding and manipulating JSON data."},
+      {"libssl",
+       "OpenSSL is a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols."},
+      {"libcrypto", "General-purpose cryptography library"},
+      {"libavcodec",
+       "Generic encoding/decoding framework and contains multiple decoders and encoders for audio, video and subtitle streams, and several bitstream filters."},
```
haydenmc, line 36: Might be worth mentioning that this is provided by ffmpeg folks

lib/glimesh/credits.ex
```diff
@@ -0,0 +1,54 @@
...
+      {"spdlog", "Very fast, header-only/compiled, C++ logging library."}
+    ]
```
haydenmc, line 38: For Orchestrator we're also using Catch2 test framework.

And for http we're using https://github.com/yhirose/cpp-httplib

PR #258, opened 2020-11-15
Real Permissions Model

clone1018:
Adding example of apps permissioning using Bodyguard


clone1018:
One thing to note, this also makes it so even if you request a resource that doesn't exist it still gives you the same error message:

image

mja00 (approved):

PR #260, opened 2020-11-16
Switching chat permissions to bodyguard

clone1018:


lib/glimesh/chat/policy.ex
```diff
@@ -0,0 +1,83 @@
...
+  # Admins
+  def authorize(:ban, %User{is_admin: true}, _channel), do: true
+  def authorize(:unban, %User{is_admin: true}, _channel), do: true
+  def authorize(:short_timeout, %User{is_admin: true}, _channel), do: true
+  def authorize(:long_timeout, %User{is_admin: true}, _channel), do: true
+
+  # Channel Owners
+  def authorize(:ban, %User{id: user_id}, %Channel{user_id: channel_user_id})
+      when user_id == channel_user_id,
+      do: true
```
mja00, line 55: Just a curiosity, is there any way of checking that the user is the channel owner without doing a when? Either way this works without issues it seems, just curious if there's a better way for quick readability.

mja00 (approved):

lib/glimesh/chat/policy.ex
clone1018: You could do it in the `do` block as well, but it would return false instead if skipping the function, which means the moderators code below would never trigger.

PR #266, opened 2020-11-18
Added timestamps to chat messages 

mja00:
Added timestamps that convert to local time(if able to), and added a site wide toggle to enable or disable them in the chat settings. 

Also tweaked the assigns for the chat box. It looks like the temporary assign was causing issues with retrieving the currently displayed chat messages, so I removed it. I haven't noticed anything wrong/weird with it. However if it was doing something then I would be able to add it back and revert the timestamp toggle to an earlier system, albeit a system that is a little less clean.

### DIFF
--- a/lib/glimesh/chat.ex
+++ b/lib/glimesh/chat.ex
@@ -169,6 +169,17 @@ defmodule Glimesh.Chat do
     |> Enum.reverse()
   end
 
+  def list_chat_messages(channel, limit) do
+    Repo.all(
+      from m in ChatMessage,
+        where: m.is_visible == true and m.channel_id == ^channel.id,
+        order_by: [desc: :inserted_at],
+        limit: ^limit
+    )
+    |> Repo.preload([:user, :channel])
+    |> Enum.reverse()
+  end
+
   @doc """
   Gets a single chat_message.
 
--- a/lib/glimesh/chat.ex
+++ b/lib/glimesh/chat.ex
@@ -158,18 +158,7 @@ defmodule Glimesh.Chat do
       [%ChatMessage{}, ...]
 
   """
-  def list_chat_messages(channel) do
-    Repo.all(
-      from m in ChatMessage,
-        where: m.is_visible == true and m.channel_id == ^channel.id,
-        order_by: [desc: :inserted_at],
-        limit: 5
-    )
-    |> Repo.preload([:user, :channel])
-    |> Enum.reverse()
-  end
-
-  def list_chat_messages(channel, limit) do
+  def list_chat_messages(channel, limit \\ 5) do
     Repo.all(
       from m in ChatMessage,
         where: m.is_visible == true and m.channel_id == ^channel.id,
--- a/lib/glimesh_web/live/chat_live/index.ex
+++ b/lib/glimesh_web/live/chat_live/index.ex
@@ -39,7 +39,7 @@ defmodule GlimeshWeb.ChatLive.Index do
       |> assign(:permissions, Chat.get_moderator_permissions(channel, session["user"]))
       |> assign(:chat_messages, list_chat_messages(channel))
       |> assign(:chat_message, %ChatMessage{})
-      |> assign(:show_timestamps?, (if session["user"], do: session["user"].show_timestamps?, else: false))
+      |> assign(:show_timestamps, (if session["user"], do: session["user"].show_timestamps, else: false))
 
     {:ok, new_socket}
   end
@@ -118,15 +118,15 @@ defmodule GlimeshWeb.ChatLive.Index do
 
   @impl true
   def handle_event("toggle_timestamps", _params, socket) do
-    timestamp_state = Kernel.not(socket.assigns.show_timestamps?)
+    timestamp_state = Kernel.not(socket.assigns.show_timestamps)
     {:ok, user} =
-      Glimesh.Accounts.User.user_settings_changeset(socket.assigns.user, %{show_timestamps?: timestamp_state})
+      Glimesh.Accounts.User.user_settings_changeset(socket.assigns.user, %{show_timestamps: timestamp_state})
       |> Glimesh.Repo.update()
     {:noreply,
      socket
      |> assign(:update_action, "replace")
      |> assign(:chat_messages, list_chat_messages(socket.assigns.channel, Kernel.length(socket.assigns.chat_messages)))
-     |> assign(:show_timestamps?, timestamp_state)
+     |> assign(:show_timestamps, timestamp_state)
      |> assign(:user, user)}
   end
 end
--- a/lib/glimesh_web/live/chat_live/index.ex
+++ b/lib/glimesh_web/live/chat_live/index.ex
@@ -50,7 +50,7 @@ defmodule GlimeshWeb.ChatLive.Index do
       |> assign(:chat_messages, list_chat_messages(channel))
       |> assign(:chat_message, %ChatMessage{})
       |> assign(:show_timestamps, user_preferences.show_timestamps)
-      |> push_event("add_timestamp_to_initial_messages", %{
+      |> push_event("update_previous_messages_with_timestamp_state", %{
         show_timestamps: user_preferences.show_timestamps
       })
       |> assign(:user_preferences, user_preferences)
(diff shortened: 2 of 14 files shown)


lib/glimesh_web/live/chat_live/index.html.leex
```diff
@@ -8,17 +8,17 @@
...
             <% end %>
             <%= if Map.get(@permissions, :can_long_timeout, false)  do %>
-            <i class="fas fa-clock fa-fw" phx-click="long_timeout_user"
+            <i id="long-timeout" class="fas fa-clock fa-fw" phx-click="long_timeout_user"
                 phx-value-user="<%= chat_message.user.username %>" data-toggle="tooltip"
                 title="<%= gettext("Timeout user for 15 minutes.") %>"></i>
             <% end %>
             <%= if Map.get(@permissions, :can_ban, false)  do %>
-            <i class="fas fa-gavel fa-fw" phx-click="ban_user" phx-value-user="<%= chat_message.user.username %>"
+            <i id="ban" class="fas fa-gavel fa-fw" phx-click="ban_user" phx-value-user="<%= chat_message.user.username %>"
```
clone1018, line 21: id's are meant to be unique on the page: https://www.w3.org/TR/WCAG20-TECHS/H93.html

priv/repo/migrations/20201118162938_add_timestamps_to_chat.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glimesh.Repo.Migrations.AddTimestampsToChat do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :show_timestamps?, :boolean, default: false
```
clone1018, line 6: Does this result in having the column actually called `show_timestamps?`? If so I think we should change it, as it would need to be escaped when using it a select statement. It's also very non-traditional :P

test/glimesh/chat_test.exs
```diff
@@ -25,6 +25,21 @@ defmodule Glimesh.ChatTest do
       chat_message
     end
 
+    def multiple_chat_message_fixture(attrs \\ %{num_of_messages: 10}) do
+      channel = channel_fixture()
+      user = user_fixture()
+      num_of_messages = attrs.num_of_messages - 2
+
+      # Pretty janky way of creating multiple chat messages but it works
+      Enum.each(0..num_of_messages, fn _i -> Chat.create_chat_message(user, channel, attrs |> Enum.into(@valid_attrs)) end)
```
clone1018, line 34: You could remove line 31 entirely if you did

```suggestion
      created_messages = Enum.each(1..num_of_messages, fn _i -> Chat.create_chat_message(user, channel, attrs |> Enum.into(@valid_attrs)) end)
      
      {:ok, chat_message} = tl(created_messages)
```

lib/glimesh/chat.ex
```diff
@@ -169,6 +169,17 @@ defmodule Glimesh.Chat do
     |> Enum.reverse()
   end
 
+  def list_chat_messages(channel, limit) do
```
clone1018, line 172: Should we combine this into the function above and add a default limit of 5?

lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -39,8 +39,9 @@ defmodule GlimeshWeb.ChatLive.Index do
       |> assign(:permissions, Chat.get_moderator_permissions(channel, session["user"]))
       |> assign(:chat_messages, list_chat_messages(channel))
       |> assign(:chat_message, %ChatMessage{})
+      |> assign(:show_timestamps?, (if session["user"], do: session["user"].show_timestamps?, else: false))
 
-    {:ok, new_socket, temporary_assigns: [chat_messages: []]}
```
clone1018, line 43: I think this had something to do with timing out messages, can you try with two accounts and then time one out?

lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -99,9 +100,33 @@ defmodule GlimeshWeb.ChatLive.Index do
...
 
+  defp list_chat_messages(channel, limit) do
+    # Only currently used when updating elements about the chat.
+    # Just makes sure that if the limit is above 50, it's set back to 50.
+    # Otherwise you could possibly load a few hundred messages when updating elements.
+    # The last 50 should do fine without anyone noticing.
+    # If the viewer's chat currently has less than 50 messages then it just gets however many there are.
+    limit = if limit > 50, do: 50, else: limit
+    Chat.list_chat_messages(channel, limit)
+  end
```
clone1018, line 111: What do you mean by this comment? Does it mean that every time a new chat message comes in it refreshes the DOM?

lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -99,9 +100,33 @@ defmodule GlimeshWeb.ChatLive.Index do
...
 
     "--chat-bg-image: url('#{url}');"
   end
+
+  @impl true
+  def handle_event("toggle_timestamps", _params, socket) do
+    timestamp_state = Kernel.not(socket.assigns.show_timestamps?)
+    {:ok, user} =
+      Glimesh.Accounts.User.user_settings_changeset(socket.assigns.user, %{show_timestamps?: timestamp_state})
+      |> Glimesh.Repo.update()
```
clone1018, line 124: Got two build warnings here ^

lib/glimesh_web/live/chat_live/index.html.leex
clone1018: Can easily be a class :)

clone1018 (changes requested):
Awesome to get some tests around chat :D!

lib/glimesh/chat.ex
mja00: I was thinking of doing that but ended up going with this. Either way works for me.

lib/glimesh_web/live/chat_live/index.ex
mja00: Due to the channel_messages list being blank/unordered when reloading all the messages with the timestamp it caused issues. This function will grab up to the last 50 messages from the DB when toggling timestamps.

priv/repo/migrations/20201118162938_add_timestamps_to_chat.exs
mja00: Yeah it does, I'll swap it over to not include the ?

test/glimesh/chat_test.exs
mja00: Unfortunately that wouldn't work as Enum.each just returns a :ok when it is finished. I'm using that solely to loop n times to create a chat message. Unless there is a function that takes the results from the inner function and puts it into a list.

clone1018: Oops, make it an Enum.map

lib/glimesh_web/live/chat_live/index.ex
mja00: I have to reload the chat when toggling timestamps otherwise they don't show up till the viewer refreshes. Not really sure why since it _is_ a liveview.

lib/glimesh_web/live/chat_live/index.ex
mja00: Timed out 2nd account and the message deleted for both views, not sure what it did previously without it.

lib/glimesh_web/live/chat_live/index.ex
mja00: I think commit ba949c1 completely negates the need for this now as toggling the timestamps actually does what it should now. I believe the temporary assign was causing issues in the first place but it looks like how we handle timeouts is no longer effected by that not being blank.

clone1018 (approved):
Preliminary approval, want to pull it down and test it here in a bit :)

clone1018:
It does look like on every single new message, it is updating the DOM of every existing message.

image

Going to see if this is happening on production or not.

clone1018:
Yup, confirmed production still uses `append` correctly:

image

We'll need to figure out how to get this same effect. Can talk about it in Discord too.

clone1018:
Whenever I send a chat message that overflows, it seems like it's cut off with the added timestamps:

image

Also disabling timestamps doesn't remove previous ones:

image

clone1018:
For reference, there should be a light margin under a newly posted chat message that splits it out from the chat message form:

image

mja00:
> Also disabling timestamps doesn't remove previous ones:

Default behavior is to not update previous messages with the new state, the last message updating is actually a bug. 

For the weirdness with cutoff messages I'll take a look.

clone1018 (approved):

PR #286, opened 2020-12-07
Made the toggle button work for logged out viewers

mja00:
When the params map is empty(logged out user), the toggle_timestamp event will only trigger the JS side of it, which will toggle the timestamps for that page. They will have to either log in to persist through refreshes/different chats or keep toggling it back on.

### DIFF
--- a/lib/glimesh_web/live/chat_live/index.ex
+++ b/lib/glimesh_web/live/chat_live/index.ex
@@ -92,7 +92,7 @@ defmodule GlimeshWeb.ChatLive.Index do
   end
 
   @impl true
-  def handle_event("toggle_timestamps", params, socket) when params == %{} do
+  def handle_event("toggle_timestamps", params, socket) when map_size(params) == 0 do
     timestamp_state = Kernel.not(socket.assigns.show_timestamps)
     {:noreply,
      socket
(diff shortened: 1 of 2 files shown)


lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -91,6 +91,19 @@ defmodule GlimeshWeb.ChatLive.Index do
     {:noreply, assign(socket, :chat_messages, list_chat_messages(socket.assigns.channel))}
   end
 
+  @impl true
+  def handle_event("toggle_timestamps", params, socket) when params == %{} do
```
clone1018, line 95:
```suggestion
  def handle_event("toggle_timestamps", %{}, socket) do
```

clone1018 (approved):

lib/glimesh_web/live/chat_live/index.ex
clone1018: If you want to be more elixiry

mja00:
So the reason the `when` remains is due to the fact that the `%{}` pattern matches to every map when in a function. The when allows me to check the `params` map and see if it is empty while letting the other map pass through to the function below.

PR #294, opened 2020-12-12
New users list design + search + scrolling

clone1018:
image

### DIFF
--- a/lib/glimesh/accounts.ex
+++ b/lib/glimesh/accounts.ex
@@ -14,6 +14,11 @@ defmodule Glimesh.Accounts do
     Repo.all(from u in User, where: not is_nil(u.confirmed_at))
   end
 
+  def search_users(query) do
+    like = "%#{query}%"
+    Repo.all(from u in User, where: ilike(u.username, ^like))
+  end
+
   def list_admins do
     Repo.all(from u in User, where: u.is_admin == true)
   end
--- a/lib/glimesh/accounts.ex
+++ b/lib/glimesh/accounts.ex
@@ -11,12 +11,19 @@ defmodule Glimesh.Accounts do
   ## Database getters
 
   def list_users do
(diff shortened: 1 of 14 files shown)


lib/glimesh_web/uploaders/avatar.ex
```diff
@@ -31,7 +31,8 @@ defmodule Glimesh.Avatar do
   end
 
   # Provide a default URL if there hasn't been a file uploaded
-  def default_url(_version, _scope) do
-    "/images/200x200.jpg"
+  def default_url(_version, scope) do
+    hash = :crypto.hash(:md5, String.downcase(scope.email)) |> Base.encode16(case: :lower)
+    "https://www.gravatar.com/avatar/#{hash}?s=200&d=robohash"
```
clone1018, line 36: Also implements Gravatar for users without avatars :)

lib/glimesh/accounts.ex
```diff
@@ -11,7 +11,19 @@ defmodule Glimesh.Accounts do
   ## Database getters
 
   def list_users do
-    Repo.all(from u in User, where: not is_nil(u.confirmed_at))
+    Repo.all(from(u in User))
```
mja00, line 14:
```suggestion
    Repo.all(from u in User, where: u.is_banned == false)
```
This'll make it skip over any globally banned users

lib/glimesh/accounts.ex
```diff
@@ -11,7 +11,19 @@ defmodule Glimesh.Accounts do
...
+  end
+
+  def search_users(query, current_page, per_page) do
+    like = "%#{query}%"
+
+    Repo.all(
+      from u in User,
+        where: ilike(u.username, ^like),
```
mja00, line 22:
```suggestion
        where: ilike(u.username, ^like),
        where: u.is_banned == false,
```

mja00 (changes requested):
Just two things to ignore banned users from the user listing

PR #300, opened 2020-12-14
Don't send every chat keystroke to the server 

clone1018:
and update the ChatMessage properly after saving

### DIFF
--- a/lib/glimesh/chat.ex
+++ b/lib/glimesh/chat.ex
@@ -199,7 +199,11 @@ defmodule Glimesh.Chat do
   end
 
   def empty_chat_message do
-    ChatMessage.changeset(%ChatMessage{}, %{})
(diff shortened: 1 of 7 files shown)


lib/glimesh/chat.ex
```diff
@@ -199,7 +199,11 @@ defmodule Glimesh.Chat do
   end
 
   def empty_chat_message do
-    ChatMessage.changeset(%ChatMessage{}, %{})
+    ChatMessage.changeset(%ChatMessage{}, %{
+      # Ensures that the ChatMessage is always replaced in the DOM, even if the message content doesn't change.
+      # Specifically used when posting a new message in chat.
+      fake_now_property: NaiveDateTime.utc_now()
+    })
```
clone1018, line 206: This is the real change

mja00 (approved):

Issue #228, opened 2020-10-13
Stream title need to be vertically centered above stream.


mja00:
It looks off and make me vomit. 
![image](https://user-images.githubusercontent.com/11252574/95925293-6cadaf00-0d87-11eb-881d-83f14d959e80.png)

PR #323, opened 2020-12-26
fix channel title display

Cykotiq:
should fix issue #228

### DIFF
--- a/assets/css/glimesh/components/stream.scss
+++ b/assets/css/glimesh/components/stream.scss
@@ -6,12 +6,17 @@
 }
 
 #video-column {
+    .col-8 {
+        align-self: center;
+    }
+}
+
+#video-column {
     h5 {
-        position: relative;
-        top: 6px;
-        overflow:hidden;
+        margin-bottom: 0;
+       /* overflow:hidden;
         white-space:nowrap;
-        text-overflow:ellipsis;
+        text-overflow:ellipsis; */
     }
 }
 
--- a/assets/css/glimesh/components/stream.scss
+++ b/assets/css/glimesh/components/stream.scss
@@ -9,6 +9,9 @@
     h5 {
         position: relative;
         top: 6px;
+        overflow:hidden;
+        white-space:nowrap;
+        text-overflow:ellipsis;
     }
 }
 
--- a/assets/css/glimesh/components/stream.scss
+++ b/assets/css/glimesh/components/stream.scss
@@ -5,6 +5,13 @@
     }
 }
 
+#video-column {
+    h5 {
+        position: relative;
+        top: 6px;
+    }
+}
+
 @media (min-width: 992px) {
     #chat-column {
         #chat {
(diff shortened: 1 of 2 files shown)


dori4n:
From [Discord](https://discord.com/channels/727111257953402891/727484476313305109/792290962420400148): instead of using positioning to move the text, you'd do better with setting the line height like
```css
#video-column .card-header h5
{
    line-height: 1.9; /* that'd currently result in 38px, the parent element is actually dynamically sized to 38.75px at the moment, but, good enough (setting 1.9375 doesn't work), better to fix that other sizing issue with the buttons on the right */
    margin-bottom: 0; /* avoid overflow pushing the rest of the content (video player) down */
}
```

Cykotiq:
> From [Discord](https://discord.com/channels/727111257953402891/727484476313305109/792290962420400148): instead of using positioning to move the text, you'd do better with setting the line height like
> 
> ```css
> #video-column .card-header h5
> {
>     line-height: 1.9; /* that'd currently result in 38px, the parent element is actually dynamically sized to 38.75px at the moment, but, good enough (setting 1.9375 doesn't work), better to fix that other sizing issue with the buttons on the right */
>     margin-bottom: 0; /* avoid overflow pushing the rest of the content (video player) down */
> }
> ```

I just tried it and it adds unnecessary spacing when the line does wrap and it shifts the video player downward more than necessary. I'm trying to mess with the dynamically sized container as little as possible

Cykotiq:
ultimately the whole channel title area needs to be redone or resized for mobile as well

dori4n:
> the line-height method also causes it to highlight the whole line area, rather than just the text.

Yes, because, while there may not be those types of characters present at the moment, this is where characters can extend to. There are different schools about where the center is, with the two camps being mostly divided about centering on the midpoint between x-height and baseline (most common) or using the midpoint between cap-height and the lowest point of the tail (the line height). In good typefaces those are usually the same, so it doesn't matter. Web browsers will already do this properly, so you should just let the browser align the text instead of doing it yourself. For all the fun, see https://www.supremo.co.uk/typeterms/ 😉

assets/css/glimesh/components/stream.scss
```diff
@@ -5,6 +5,16 @@
     }
 }
 
+#video-column {
+    h5 {
+        position: relative;
+        top: 6px;
```
clone1018, line 11: I'm not sure I'm in love with bumping this up an arbitrary pixel amount. We should be able to remove the margin from the h5 and the vertical align it inside of it's container.

Cykotiq: it's not bumping it up, its bumping it down an additional 6px from the top, which is fine unless another element forces the container to resize. I think flexbox `align-items: center;` may be the only way to vertically center a container who's height can change, but I'm not sure how to do that without screwing up the whole

Cykotiq: ok so I think it needs to be
```
#video-column {
    .col-8{
	align-self:center;
    }
}

h5 {
	margin-bottom:0;
}
```
do we still want it to stay on one line and cut off the end with "..."?

clone1018:
@Cykotiq I ended up coming back to this work, and copying it to https://github.com/Glimesh/glimesh.tv/pull/399

Thanks for figuring this out!

PR #330, opened 2020-12-29
Adding chat parts

clone1018:


clone1018:
Get outta here trash PR!

Issue #309, opened 2020-12-21
Expand list of OAuth 2.0 scopes to allow more fine grained access control for apps


dori4n:
From [Discord](https://discord.com/channels/727111257953402891/727484476313305109/790408653756563476):

Can I suggest adding scopes (mostly chat related) such that the following must all be authorized separately, per app and user combination:
* acting as the user on channels which are not their own
* connecting to chat (presence)
* sending a message
* deleting a message
* purging multiple messages of a user
* clearing all messages
* assigning roles (moderator, are there others a channel owner can set, yet?)
* timing out a user
* banning a user
* globally timing out a user (staff only, but still, who knows, they might behave like regular users and you might want it available)
* globally banning a user from chatting (same as above)

Anything else that modifies something or accesses something non-public, as soon as you think about making it available via the API or add a relevant new feature.

wolfcomp:
>  there is some things everyone should be aware before we start adding more scopes and that is all current existing applications would not be able to use these unless its also added to the SQL table of that application OR the entire oauth2 lib that is being used is ripped out and create our own implementation that does not have this limitation
taken from discord reply from me.

Cykotiq:
I would add to your list a permission to read subscriber/subscribee info - it's nobody's business how much money a streamer makes. Since the sub profits are well known, it's just a simple calculation to know the exact amount as streamer makes.

> > there is some things everyone should be aware before we start adding more scopes and that is all current existing applications would not be able to use these unless its also added to the SQL table of that application OR the entire oauth2 lib that is being used is ripped out and create our own implementation that does not have this limitation
> 
> taken from a discord reply from me.

I would not automatically add the permissions, as that goes against the whole idea of restricting access to anyone who has not explicitly allowed it. We can just send out an email to all 3rd party devs letting them know they will need to re-auth their users to make use of the newly restricted scopes.  Dev's should not be automatically given access to scopes they don't necessarily use. We have not even launched yet, and even once we do, we'll still be in alpha - these changes should be expected as the API is further developed

dori4n:
> I would add to your list a permission to read subscriber/subscribee info/events - it's nobody's business how much money a streamer makes.

[...]

PR #332, opened 2020-12-30
Fix OAuth2 to not break whenever we add more scopes for existing applications

wolfcomp:
This semi blocks #309 from being a realization without a lot of prep work in the SQL tables for all existing applications

### DIFF
--- a/lib/glimesh/oauth/oauth_access_grant.ex
+++ b/lib/glimesh/oauth/oauth_access_grant.ex
@@ -4,7 +4,14 @@ defmodule Glimesh.OauthAccessGrants.OauthAccessGrant do
   use ExOauth2Provider.AccessGrants.AccessGrant, otp_app: :glimesh
 
   schema "oauth_access_grants" do
-    access_grant_fields()
+    field :token, :string, null: false
+    field :expires_in, :integer, null: false
+    field :redirect_uri, :string, null: false
+    field :revoked_at, :utc_datetime
+    field :scopes, :string
+
+    belongs_to :resource_owner, Glimesh.Accounts.User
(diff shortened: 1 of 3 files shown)


Cykotiq:
I feel VERY STRONGLY that existing apps should not get new scopes grandfathered in. We can't just assume users are ok with the new permissions - they may not have been aware they opted into it in the first place (it's one short line each). Maybe they didn't know those endpoints were a part of the API. What if we add new endpoint - we can' just assume users are ok with it. Some of these scopes are being added because the previous scopes were not restrictive enough, so overriding the user's decision defeats the purpose, especially if the app doesn't actually NEED that permission.

Try to think of each scope/permission as a checkbox that the user has the ability to check or uncheck before clicking accept (I would love this BTW). You shouldn't just assume the user is ok with the new scopes being added. Case in point: the scope for subscriber details. I don't believe this should have ever been under the "public" scope, but if users see they are granting access to that info they may not be ok with it. Users don't know what the API is capable of accessing specifically without reading the API docs (which normal users will not do).

Requiring apps to re-auth is not a big deal, and honestly it should be EXPECTED since we are still pre-alpha, and even beyond then until beta at least. It's not worth writing extra code over, much less code that bypasses the user's decisions

wolfcomp:
The part that would break that this is going to fix is not the user accepted scopes but what the actuall application can use when creating new tokens for approval of the user.

So in the sence current connected apps with approved tokens from users will NOT get the new scopes if their added BUT if there was a new authrorization request and the user accepted this they would get access to the new scopes if these were suplied in the scope object.

wolfcomp:
Continues in the pull request mentioned above

PR #348, opened 2021-01-04
Adding new load balancer cookie

clone1018:


clone1018:
@Heiwa1580 This look good to you?

Heiwa1580 (commented):
Instead of functional, required cookies. I would say

Glimesh uses only strictly necessary cookies for the website. We do not allow any marketing, statistic, or preference cookies. 

https://gdpr.eu/cookies

> To comply with the regulations governing cookies under the GDPR and the ePrivacy Directive you must:
> 
> Receive users’ consent before you use any cookies **except** strictly necessary cookies.
> Provide accurate and specific information about the data each cookie tracks and its purpose in plain language before consent is received.
> Document and store consent received from users.
> Allow users to access your service even if they refuse to allow the use of certain cookies
> Make it as easy for users to withdraw their consent as it was for them to give their consent in the first place.

clone1018:
Thanks, adjusted!

PR #352, opened 2021-01-07
Added fixes and new features to the GCT dash

mja00:
- Fixed `View on GCT` button on profile pages
- Componentized and added to GCT dash
  - Profile settings page
  - Channel settings page
- New social formats show on GCT dash due to above
- Added better searching for users and channels.
  - User search can be done with:
    - Username
    - Email
    - User ID
  - Channel search can be done with:
    - Username
    - Channel ID

PR #353, opened 2021-01-07
Added rotate stream key button

mja00:
The button will trigger the Streams.rotate_stream_key function and then show the new stream key. Button also has a 5-second throttle on it so people can't spam it. 

Must be merged after #352 as it's based on the new LiveView component that was added for the channel settings page.


mja00:
The above commits are ugly AF but it's okay. :)

mja00:
Replaced by #354 as this branch became weird.

PR #364, opened 2021-01-10
Fixes applications without a homepage breaking the authorized page

mja00:


clone1018 (approved):
Approved pending tests

PR #372, opened 2021-01-11
Added the channel ID to prefix the stream key

### LINKED ISSUE #370: Stream key shown to user should include channel prefix
Current format:
`<stream_key>`

Correct format: 
`<channel_id>-<stream_key>`

Locations:
- [ ] API
- [ ] UI on stream page

mja00:
Fixed #370 

Instead of storing just the generated stream key by itself in the DB, the channel ID is prefixed to it with a dash as a seperator. 

The new saved format is now `<channel_id>-<stream_key>` as outlined in the linked issue.

### DIFF
--- a/lib/glimesh/streams.ex
+++ b/lib/glimesh/streams.ex
@@ -49,11 +49,13 @@ defmodule Glimesh.Streams do
 
   def create_channel(%User{} = user, attrs \\ %{category_id: Enum.at(list_categories(), 0).id}) do
     with :ok <- Bodyguard.permit(__MODULE__, :create_channel, user) do
-      %Channel{
+      {:ok, channel } = %Channel{
         user: user
       }
       |> Channel.create_changeset(attrs)
       |> Repo.insert()
+
+      rotate_stream_key(user, channel) # Has to be done to generate a stream key for the channel
     end
   end
 
@@ -79,7 +81,7 @@ defmodule Glimesh.Streams do
     with :ok <- Bodyguard.permit(__MODULE__, :update_channel, user, channel) do
       channel
       |> change_channel()
-      |> Channel.stream_key_changeset()
+      |> Channel.stream_key_changeset(channel.id)
       |> Repo.update()
     end
   end
--- a/lib/glimesh/streams/channel.ex
+++ b/lib/glimesh/streams/channel.ex
@@ -37,7 +37,6 @@ defmodule Glimesh.Streams.Channel do
     channel
     |> changeset(attrs)
     |> put_change(:status, "offline")
-    |> put_change(:stream_key, generate_stream_key())
   end
 
   def start_changeset(channel, attrs \\ %{}) do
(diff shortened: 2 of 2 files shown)


clone1018:
Apparently Janus authenticates using the original stream key via the API. This change would break stream authentication, what should we do here?

mja00:
Could we split the string by `-` and return the 2nd half through the API?

haydenmc:
Feels counter-intuitive to me to persist the channel ID portion of the stream key, since that's duplicating information that's already available in another column.

Is there a reason you can't store the stream key by itself and just present it to the user as `[channelid]-[streamkey]`?

clone1018:
I think my problem is with it being called the same thing, but being used in two different ways. 

For Janus (using the api) streamKey which SHOULD NOT not have the channel ID.
For a user (using the api) streamKey SHOULD have the channel ID.
For a user (using the web) streamKey SHOULD have the channel ID.

Maybe it's as simple as calling it something else? Channel Key? Code thing?

mja00:
Possible store another form of the key for only Janus in the DB that it can pull from that just has the key?

haydenmc:
The channel ID portion of the 'key' isn't really the 'key' I suppose - it's the username equivalent to a username/password form.

Maybe we can call the whole thing a stream 'token' or something.

`[channelid]` - The channel ID
`[streamkey]` - The stream key
`[channelid]-[streamkey]` - The stream token

I don't think we should be duplicating any of these values and storing them in different places/formats - I feel that's just asking for integrity failures if any of them end up changing out from under you.

clone1018:
I agree to not storing it in it's combined state, but I do want to make sure we're using the right names for the values. In OBS the value is called Stream Key, and I want to make sure API users and website users don't get confused when they can access a value called `streamKey` that doesn't actually contain the entire stream key they need. 

> `[channelid]` - The channel ID
> `[streamkey]` - The stream key
> `[channelid]-[streamkey]` - The stream token

I'll agree to this if we can instead do:
`[channelid]` - The channel ID
`[streamkey]` - The stream passcode / password / identifier / some other word
`[channelid]-[streamkey]` - The stream key

haydenmc:
> [streamkey] - The stream passcode / password / identifier / some other word

It's rough because that value is actually an HMAC 'key' in the literal sense, so by calling it something else we risk confusing ourselves (which I guess is better than confusing the user).

Hey, maybe just call it hmackey and the combined value streamkey.

clone1018:
> Hey, maybe just call it hmackey and the combined value streamkey.

I'm happy with that.

mja00:
So the consensus is: 
- Revert changes in this PR
- Display on the frontend a concatenation of the channel ID and the key
- API for users does the same as above
- Janus just gets the key
?

clone1018:
I think it's uh...
0. Yeet PR
1. Rename existing field to hmackey
2. Adjust API to return both hmackey & stream key
3. Concatenate the values for the frontend (or better yet created a shared function)
5. Ignore me and do what feels right

mja00:
PR yote

PR #373, opened 2021-01-14
Update privacy.html.md to include support portal cookie details

moltenllama:
Added Support Portal cookie details.


clone1018:
Looks good, thanks!

PR #378, opened 2021-01-14
Add text-justify class to align text

EmberShadow:


clone1018:
Looks good, thanks again!

PR #379, opened 2021-01-14
:bug: Allow drift between 2FA device clock and backend

### LINKED ISSUE #360: 2FA Auth timing doesn't always work
> 2FA is too strict so code won't work unless it's immediately fresh for some users. Can make it annoying to sign in

shrink:
Closes #360

According to [RFC 6238](https://tools.ietf.org/html/rfc6238):

>  Because of possible clock drifts between a client and a validation
   server, we **RECOMMEND that the validator be set with a specific limit
   to the number of time steps a prover can be "out of synch" before
   being rejected.**
>
>   This limit can be set both forward and backward from the calculated
   time step on receipt of the OTP value.  If the time step is
   30 seconds as recommended, and the validator is set to only accept
   two time steps backward, then the maximum elapsed time drift would be
   around 89 seconds, i.e., 29 seconds in the calculated time step and
   60 seconds for two backward time steps.
>
>   This would mean the validator could perform a validation against the
   current time and then two further validations for each backward step
   (for a total of 3 validations).  Upon successful validation, the
   validation server can record the detected clock drift for the token
   in terms of the number of time steps.  When a new OTP is received
   after this step, the validator can validate the OTP with the current
   timestamp adjusted with the recorded number of time-step clock drifts
   for the token.

This change adds an additional `step` parameter when using `validate_pin`. The number of steps is the number of _additional_ backward steps, the `validate_pin` method will **always** perform a check of the pin against the current time.

### DIFF
--- a/lib/glimesh/tfa.ex
+++ b/lib/glimesh/tfa.ex
@@ -23,9 +23,10 @@ defmodule Glimesh.Tfa do
     EQRCode.png(input, width: 355, background_color: <<255, 255, 255>>, color: <<0, 0, 0>>)
   end
 
-  defp generate_hmac(secret, period) do
+  defp generate_hmac(secret, period, seconds_ago) do
     moving_factor =
       DateTime.utc_now()
+      |> DateTime.add(seconds_ago * -1)
       |> DateTime.to_unix()
       |> Integer.floor_div(period)
       |> Integer.to_string(16)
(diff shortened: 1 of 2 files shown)


clone1018:
Thanks for the work! Tested it and it seemed to handle a potential drift well. Nice tests :D!

PR #382, opened 2021-01-15
📐 Left align "Community Driven" homepage header

haydenmc:
This change left aligns the "Community Driven" header on the home page to avoid a ragged left edge, which is generally [frowned upon](https://mrwweb.com/no-justification-dont-use-right-center-and-full-justification-on-the-web/) (with LTR languages at least) due to the negative impact it has on reading speed and comprehension, plus the inconsistent aesthetic.

## Before:
Before change, with right-justified text

## After:
After change, with left-justified text


clone1018:
Looks good, not sure exactly why the tests are failing...

haydenmc:
Ah, I was targeting `master` instead of `dev`. Rebasing...

PR #384, opened 2021-01-16
Temporary fix for issue Glimesh#383

Drehd:
# Before:
image

# After:
image


clone1018:
Couple of test errors here @Drehd. Should be simple text matches :D

Drehd:
Noticed on even smaller displays the buttons still would be one on top of the other, so I made the button resize to prevent that in the last commit.

assets/css/glimesh/components/stream.scss
```diff
@@ -24,4 +24,37 @@
...
     }
+}
+
+@media (max-width: 768px) {
+    .btn-responsive {
+        padding: 2px 4px;
+        line-height: 1;
+        border-radius: 3px;
+    }
+    .f_txt {
```
clone1018, line 35: What do you think about replacing these with a common class you can add to each element you want to hide? 

Optionally Bootstrap provides these classes we can use to achieve the same effect with the Bootstrap best practices included: https://getbootstrap.com/docs/4.6/utilities/display/#hiding-elements

clone1018:
Thanks for the work, I tested this out locally and it works beautifully! Just have a comment on the CSS.

assets/css/glimesh/components/stream.scss
Drehd: That looks way better, going to try that now, thank you!

clone1018: Looks awesome!

clone1018 (approved):

PR #388, opened 2021-01-19
Fix some bottom spacing on the about/team page

wolfcomp:


lib/glimesh_web/templates/about/team.html.eex
```diff
@@ -6,7 +6,7 @@
 
 <div class="card">
     <div class="card-body">
-        <p>
+        <p style="margin:0;">
```
clone1018, line 9:
```suggestion
        <p class="m-0">
```

clone1018 (changes requested):

lib/glimesh_web/templates/about/team.html.eex
mja00: +1

PR #389, opened 2021-01-20
Change the From header in emails to properly mention "Glimesh"

LittleToonCat:
With certain email providers (such as Gmail), emails sent through the Glimesh website would make it say it was from "support" instead of Glimesh.  That is because the From header of these emails only contains the email address, causing it to show the "user" of the address instead.

This PR fixes the situation by properly formatting the header to include the "Glimesh" name.

Before:
image
After:
image


clone1018:
Looks great, thanks!

PR #393, opened 2021-01-20
This will prevent tables that are larger than the page from overflowi…

EmberShadow:
…ng and will add a scrollbar as needed


clone1018:
Looks good, thanks!

PR #412, opened 2021-01-27
Update badwords.exs

EmberShadow:


clone1018:
Thank you for your service.

PR #428, opened 2021-02-03
Adding API Logging functionality

clone1018:
Adds a simple API logger so we can understand how people are using the API. Shouldn't leak any query or mutation details.


mja00 (approved):
send this straight to prod k thx

PR #435, opened 2021-02-08
Removing Paddy

Heiwa1580:
Just a change to remove Paddy from the list.


Heiwa1580:
The reason I think we don't need to have this added is I don't think that the word warrants being blocked. It's not an inherently racial term and is very multifaceted. The name can be used as a shortened version of Patrick and many Irish people use it like American's use Yankee.

PR #439, opened 2021-02-10
Adding cross-border payouts

clone1018:


clone1018:
Should be ready for review, but I'm going to double check it tomorrow.

.github/workflows/elixir.yml
```diff
@@ -27,22 +27,31 @@ jobs:
...
       uses: actions/cache@v2
       with:
         path: deps
         key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
         restore-keys: ${{ runner.os }}-mix-
+
     - name: Install dependencies
       run: mix deps.get
+
+    - name: Start stripe-mock
```
clone1018, line 47: The stripe mock we were using before only had endpoints for like 3 Stripe functions. This package is a full distribution straight from Stripe.

The downside right now is that means you'll need `stripe-mock` running locally or the payment tests wont pass.

lib/glimesh_web/controllers/gct_controller.ex
```diff
@@ -61,7 +61,7 @@ defmodule GlimeshWeb.GctController do
           conn,
           "lookup_user.html",
           user: user,
-          payout_history: Payments.list_payout_history(user),
+          payout_history: [],
```
clone1018, line 64: Disabled this for now because it's not accurate.

lib/glimesh_web/templates/user_payments/setup.html.eex
```diff
@@ -0,0 +1,45 @@
+<div class="container">
+    <h2 class="mt-4"><%= gettext("Setup Glimesh Payouts") %></h2>
+
+    <%= if @can_payments do %>
```
clone1018, line 4: I'm not gettexting a lot of this page yet because it is nothing near the final text.

mix.exs
```diff
@@ -81,11 +80,11 @@ defmodule Glimesh.MixProject do
       {:waffle_ecto, "~> 0.0.9"},
       {:ex_aws, "~> 2.1.2"},
       {:ex_aws_s3, "~> 2.0"},
-      {:hackney, "~> 1.9"},
+      {:hackney, "~> 1.17"},
```
clone1018, line 83: Required update due to https://github.com/benoitc/hackney/issues/664

PR #441, opened 2021-02-11
Chat height adjust fix

wolfcomp:
To make viewing and chatting a bit easier on large laptop screens running 16:9 aspect

Before:
before

After:
after


wolfcomp:
On IPad 10.5":

Before:
before

After:
after

wolfcomp:
On iPhone 11 Portrait:

Before:
before

After:
after

clone1018:
Responded in Discord as well, but this new update works perfect on all normal laptop + desktop screen resolutions! Woo!

One issue I noticed is on mobile when the chat has no contents: 
image
Which should just be a min-height or something away! Once we get that fixed I think we're good to go.

wolfcomp:
idk what happened with coveralls and why it complains about coverage decreased as i have not changed any of the files it complains on

assets/css/glimesh/components/chat.scss
```diff
@@ -89,4 +89,18 @@
     .chat-form {
         flex: 0 0;
     }
+}
+
+.chat-space-fill {
```
clone1018, line 94: I played around with this PR for awhile, and I noticed that removing this .chat-space-fill and the associated image in `lib/glimesh_web/live/user_live/stream.html.leex` have no effect on any browser size that I tested.

Tested:
1920×1080
1366×768
1440×900
1536×864
2560×1440
1680×1050
1280×720
1280×800
360×640
1600×900

What exactly is the selector for?

clone1018: To clarify, even when I removed `.chat-space-fill` the code that was changed in `stream.scss` was **perfect** on all of those resolutions.

wolfcomp: its to have something take up the space under the chat as it is somewhat floating and this would be only seen on laptop screens as these are the only one that has the right width and height ratio to activate this css styling ref in the picture
glim

clone1018: Can you provide an example of a resolution I can test this on?

wolfcomp: 1440x732

wolfcomp: this is the most common resolution for web browsers on large laptop screens with the favorites bar enabled

clone1018: Okay, I confirmed on that resolution that this `.chat-space-fill` is not requried for us to get the functionality of the PR.

.chat-space-fill existing
image


.chat-space-fill completely deleted
image

Can we remove this selector and the img?

wolfcomp: Can do that was more to have something there.

clone1018: I feel like we're just going to get people reporting that there's a glimdrop there :P

wolfcomp: XD

PR #451, opened 2021-02-13
Fix followers page

clone1018:
Broken when I refactored the stream list for performance.


clone1018:
Should add tests for these eventually lol...

PR #479, opened 2021-02-22
Change avatar url to be of user id instead of username

wolfcomp:


priv/repo/migrations/20210222011540_fix_avatar_url_paths.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Glimesh.Repo.Migrations.FixAvatarUrlPaths do
+  use Ecto.Migration
+
+  def change do
+    Enum.each(Glimesh.Accounts.list_users(), fn user ->
+      if user.avatar do
+        avatar_url = Glimesh.Avatar.url({user.avatar, user}, :original_x2)
+        if String.starts_with?(avatar_url, "https") do
+          {:ok, resp} = :httpc.request(:get, {avatar_url, []}, [], [body_format: :binary])
```
wolfcomp, line 9: This will not work as avatar_url is not of type charlist the fix is on line 7 to include also `|> to_charlist()`

PR #481, opened 2021-02-23
Adding tag search & creating better mock environment

clone1018:
Adds a tag search functionality to the category pages:
image

Also handles updating the browser URL with ?tags[]=one&tags[]=two so we can save the users state a bit.

Cleans up the seed environment (by only using categories), and introduces a new `priv/repo/mocks` directory you can use for creating some fake data.


mja00 (approved):
:)

PR #483, opened 2021-02-23
Stream Shutdown for GCT

clone1018:


haydenmc:
Per my comment on Discord, eventually some more precise error codes would be nice, so Janus-FTL can determine how to best handle them (retry/end stream/etc.)

But this is fine for now™ 😁

haydenmc (approved):

mja00 (approved):

PR #487, opened 2021-02-24
Add try statement on link display code to never break on invalid links

wolfcomp:
so team meeting incident on 02/24/2021 never happens again

PR #488, opened 2021-02-24
Prevent chat crashing the whole live view

clone1018:


clone1018:
Implements https://github.com/Glimesh/glimesh.tv/pull/487 and adds some glitter on top.

clone1018:
Also adds a new 500 page that looks like: 
image

PR #498, opened 2021-02-26
Allowing users to recreate their channel

clone1018:


lib/glimesh/channel_lookups.ex
```diff
@@ -14,10 +14,11 @@ defmodule Glimesh.ChannelLookups do
 
   def list_channels(wheres \\ []) do
     Repo.all(
-      from c in Channel,
+      from(c in Channel,
```
clone1018, line 17: I cannot figure out what is doing this formatting....

lib/glimesh/channel_lookups.ex
```diff
@@ -119,19 +127,21 @@ defmodule Glimesh.ChannelLookups do
...
-  def get_channel_for_user(user) do
+  def get_channel_for_user(user, include_inaccessible \\ false) do
     Repo.one(
-      from c in Channel,
+      from(c in Channel,
         join: u in User,
         on: c.user_id == u.id,
         where: u.id == ^user.id,
-        where: c.inaccessible == false
+        where: c.inaccessible == ^include_inaccessible
```
mja00, line 143: So when `include_inaccessible` is set to true it will **only** find rows where `inaccessible` is true. I ran into this issue when including banned users in searches. I couldn't find an easy way to conditionally adding a new where clause.

mja00 (changes requested):
Just that one thing. Look at what I did for the get user functions, it's not pretty but it works, somewhat at least.

lib/glimesh/channel_lookups.ex
clone1018: Just made a new function to not introduce confusion.

mja00 (approved):
Looks good to me now

PR #499, opened 2021-02-26
Translation updates as of 02/26/2021

mja00:
Waiting on the proper names for the Chinese variants in their respective languages.


mja00:
Merge #500 first

PR #501, opened 2021-02-27
Fix Stream title height to match dropdown height

ACPixel:


assets/css/glimesh/components/stream.scss
```diff
@@ -33,4 +33,8 @@
         line-height: 1;
         border-radius: 3px;
     }
-}
\ No newline at end of file
+}
+
+input#channel_title {
+ height: 45px;
```
wolfcomp, line 39: Make it 4 whitespaces and not 1

PR #505, opened 2021-02-27
Open profile & rules links in new tab

### LINKED ISSUE #335: Links on user profiles should open in new window

clone1018:
Fixes #335﻿


clone1018:
This should actually always work, browsers are good at having multiple html attributes.

mja00 (approved):

PR #506, opened 2021-02-27
Optimizing image upload for DO

clone1018:
This should do three major things:

1. Ensure DO uploads process within 4 seconds (before erroring gracefully to the user or API)
2. Upload images with correct content type
3. Cache images forever, until Waffle cache busts them﻿

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -43,7 +43,8 @@ config :glimesh,
 
 config :waffle,
   storage: Waffle.Storage.Local,
-  storage_dir: "uploads"
+  storage_dir: "uploads",
+  version_timeout: 4_000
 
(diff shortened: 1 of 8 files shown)


config/config.exs
```diff
@@ -43,7 +43,8 @@ config :glimesh,
 
 config :waffle,
   storage: Waffle.Storage.Local,
-  storage_dir: "uploads"
+  storage_dir: "uploads",
+  version_timeout: 4_000
```
clone1018, line 47: The driver for 4 seconds is that the default write timeout for the HTTP lib Glimesh/janus-ftl-player is 5 seconds.

mja00 (approved):

PR #508, opened 2021-02-28
Some improvements to the api.

wolfcomp:
- Added in so arguments can be converted to ids instead of using names.
- Added `isMod` field for easier checks for bot makers to make mod only commands
- Added pagination
- Added `/apinext` for current changes these should be out for 2 weeks minimum before loaded over to `/api` to give people time to change to new queries

Places with pagination:
```
followers
channels
channel{
  chatMessages
  bans
  moderationLogs
  moderators
}
```

All pagination except chatMessages has a limit of 1000 currently set where as chatMessages only has 100 as a max


wolfcomp:
Will fix tests after alpha is live

lib/glimesh/graphql/schema.ex
```diff
@@ -4,6 +4,7 @@ defmodule Glimesh.Schema do
   """
 
   use Absinthe.Schema
+  use Absinthe.Relay.Schema, :modern
```
clone1018, line 7: Is this needed? I think we may only need it on the new lib/glimesh/api/schema.ex

clone1018:
Okay, I went through and removed the files with only line ending changes and I also refactored the functions that return queries to be prefixed with `query_` and added back their `list_` or `get_` counterparts. These changes were done to ensure our queries are happening inside the contexts instead of inside controllers. 

I think this is ready for a final pass of review! Super excited to get this new, much more performant, and scalable API in place. Thanks again for your hard work @wolfcomp !

clone1018:
One more thought, is the inserted_at an atomic enough field to be sorting by for the pagination? I suspect it's possible to have two timestamps be the exact same and you could miss records in the pagination.

lib/glimesh/graphql/schema.ex
wolfcomp: Might have been something I forgot to remove once I moved it to its own schema for relay and old

clone1018 (approved):

PR #516, opened 2021-03-02
Swapping preparred statements to unnamed

mja00:
In theory this fixes the error: `prepared statement "ecto_1234" does not exist` by making all prepared queries unnamed. This makes it so that it doesn't matter which backend the query goes to. 

I was following [this](https://elixirforum.com/t/can-someone-please-shed-some-light-on-this-error/33439/10) forum post and [this](https://hexdocs.pm/ecto_sql/Ecto.Adapters.Postgres.html#module-connection-options) docs page.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -41,6 +41,9 @@ config :glimesh,
   run_stream_pruner: true,
   locales: locales
 
+config :glimesh, Glimesh.Repo,
+  prepare: :unnamed
+
 config :waffle,
   storage: Waffle.Storage.Local,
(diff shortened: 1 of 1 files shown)


clone1018:
Rebased this against dev

clone1018:
Spent a bit time researching this, seems like this change + Transaction pooling mode is going to give us much better database scaling ability, while potentially impacting individual query performance. However we're seeing an issue with the former not the latter so I'll switch prod over soon.

PR #517, opened 2021-03-03
Changing the title of the sub button when you cannot sub

mja00:


mja00:
Merge #516 and then re-run the check :)

PR #521, opened 2021-03-03
Fixing various bugs

clone1018:


wolfcomp (approved):
yup seems like good fixes

PR #524, opened 2021-03-03
Various bug fixes

mja00:
Fixed various bugs around the site:
- Dark text on a dark background on the homepage when in light mode
- Dark text on a dark background on stream lists when in light mode
- Proper light mode in chat
- Discord invites now accept any format and will truncate to only the invite code
- Chat bubble for mentions is now a red border instead of a green background
- Chat now behaves when someone is timed out
- Typo on the homepage

### DIFF
--- a/assets/js/hooks/Chat.js
+++ b/assets/js/hooks/Chat.js
@@ -89,5 +89,13 @@ export default {
             this.maybeScrollToBottom(chatMessages);
         });
 
+        this.handleEvent("remove_timed_out_user_messages", (e) => {
+            let offendingUserID = e["bad_user_id"];
+            let offendingChatMessage = chatMessages.querySelectorAll(`[data-user-id=${CSS.escape(offendingUserID)}]`);
+            Array.prototype.forEach.call(offendingChatMessage, function(node) {
+                node.hidden = true; // Have to hide them otherwise the tooltip gets stuck on removing the element
+            });
+        })
+
     }
 };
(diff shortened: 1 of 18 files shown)


assets/js/hooks/Chat.js
```diff
@@ -89,5 +89,13 @@ export default {
             this.maybeScrollToBottom(chatMessages);
         });
 
+        this.handleEvent("remove_timed_out_user_messages", (e) => {
+            let offendingUserID = e["bad_user_id"];
+            let offendingChatMessage = chatMessages.querySelectorAll(`[data-user-id=${CSS.escape(offendingUserID)}]`);
+            Array.prototype.forEach.call(offendingChatMessage, function(node) {
+                node.hidden = true; // Have to hide them otherwise the tooltip gets stuck on removing the element
+            });
```
clone1018, line 97:
```suggestion
            // Have to hide them otherwise the tooltip gets stuck on removing the element
            offendingChatMessage.forEach(e => e.hidden = true);
```

clone1018: Pretty sure this new syntax would work too, looks a bit nicer.

lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -151,8 +151,7 @@ defmodule GlimeshWeb.ChatLive.Index do
 
     {:noreply,
      socket
-     |> assign(:update_action, "replace")
-     |> assign(:chat_messages, list_chat_messages(socket.assigns.channel))
+     |> push_event("remove_timed_out_user_messages", %{bad_user_id: bad_user.id})
```
clone1018, line 154: This makes me happy, such optimization!

clone1018 (changes requested):
Looks great! One super minor change

assets/js/hooks/Chat.js
mja00: Gonna commit and try it

PR #527, opened 2021-03-04
Added live tag to profile page

mja00:
Adds the live badge to the profile page when the channel is live. This goes right above the view channel button:

image

### DIFF
--- a/lib/glimesh_web/live/user_live/profile.ex
+++ b/lib/glimesh_web/live/user_live/profile.ex
@@ -20,7 +20,7 @@ defmodule GlimeshWeb.UserLive.Profile do
 
         streamer_share_text = Profile.streamer_share_text(streamer, profile_url)
         viewer_share_text = Profile.viewer_share_text(streamer, profile_url)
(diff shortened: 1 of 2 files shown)


lib/glimesh_web/live/user_live/profile.ex
```diff
@@ -20,6 +20,7 @@ defmodule GlimeshWeb.UserLive.Profile do
 
         streamer_share_text = Profile.streamer_share_text(streamer, profile_url)
         viewer_share_text = Profile.viewer_share_text(streamer, profile_url)
+        maybe_channel = if Glimesh.ChannelLookups.get_channel_for_user(streamer), do: Glimesh.ChannelLookups.get_channel_for_user(streamer), else: nil
```
clone1018, line 23:
```suggestion
        maybe_channel = if channel =  Glimesh.ChannelLookups.get_channel_for_user(streamer), do: channel, else: nil
```

Would be double as fast.

mja00: Haha I knew there was some fast way of doing it

PR #530, opened 2021-03-05
Optimizing chat timestamps

clone1018:
Previously the timestamps were slightly expensive to render, this changes it to a root class that is toggled depending on the users setting.

### DIFF
--- a/lib/glimesh_web/live/chat_live/index.ex
+++ b/lib/glimesh_web/live/chat_live/index.ex
@@ -54,9 +54,6 @@ defmodule GlimeshWeb.ChatLive.Index do
         |> assign(:chat_messages, list_chat_messages(channel))
         |> assign(:chat_message, %ChatMessage{})
         |> assign(:show_timestamps, user_preferences.show_timestamps)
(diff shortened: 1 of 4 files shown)


lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -146,18 +139,7 @@ defmodule GlimeshWeb.ChatLive.Index do
...
-        do: Accounts.get_user_preference!(socket.assigns.user).show_timestamps,
-        else: false
-
-    {:noreply,
-     socket
-     |> push_event("remove_timed_out_user_messages", %{bad_user_id: bad_user.id})
-     |> push_event("update_previous_messages_with_timestamp_state", %{
-       show_timestamps: show_timestamps
-     })}
+    {:noreply, push_event(socket, "remove_timed_out_user_messages", %{bad_user_id: bad_user.id})}
```
clone1018, line 142: The deleted code is no longer required since timing out does not re-render the DOM.

mja00 (approved):

PR #531, opened 2021-03-05
Counting semi-unique viewers

clone1018:
Not a perfect solution, but will prevent basic "view botting" by calculating a unique user hash per IP + logged in user combo. 

In the future this will actually be a count of WebRTC video consumers.﻿

### DIFF
--- a/lib/glimesh_web/endpoint.ex
+++ b/lib/glimesh_web/endpoint.ex
@@ -8,6 +8,7 @@ defmodule GlimeshWeb.Endpoint do
   @session_options [
     store: :cookie,
     key: "_glimesh_key",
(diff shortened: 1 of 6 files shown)


lib/glimesh_web/endpoint.ex
```diff
@@ -8,6 +8,7 @@ defmodule GlimeshWeb.Endpoint do
   @session_options [
     store: :cookie,
     key: "_glimesh_key",
+    same_site: "Lax",
```
clone1018, line 11: Beefs up our cookie security, this will likely log everyone out.

mja00 (approved):

PR #536, opened 2021-03-06
Adding a modal for stream editing

clone1018:
image

Replaces the existing inline editor with a modal for editing tags as well. Should also be better on Mobile.


mja00 (changes requested):
Looks like `GlimeshWeb.UserLive.Components.ChannelTitle` is missing a handle event for `hide_editing` so when a user goes to close the modal without changing anything the page freezes during the error.

PR #555, opened 2021-03-11
Fixing typos

Heiwa1580:
Fixing the command for the emotes filenames I believe would be fixed elsewhere?


clone1018:
Yup, emotes are stored here: https://github.com/Glimesh/glimesh.tv/tree/dev/assets/static/emotes

PR #567, opened 2021-03-13
Revert "Set img-avatar to be responsive inside container"

clone1018:
Reverts Glimesh/glimesh.tv#542

@omn0mn0m I thought you maybe tested this :P

image


omn0mn0m:
Whoops that's my bad, I didn't even notice it changed that as well...

clone1018:
:P

PR #568, opened 2021-03-13
Adds count all followers to API

AdamHebby:
The methods already existed, I just gave the API a way to access them.

Honestly, not a clue if this works, idk how to test it - there are no tests for similar functions (like all_followers) so I couldn't copy those. It took me ages to find similar functions in these files, I copied them and changed them slightly to what seems correct to use the new functions. Pls test!

### DIFF
--- a/lib/glimesh/graphql/schema/account_types.ex
+++ b/lib/glimesh/graphql/schema/account_types.ex
@@ -4,6 +4,7 @@ defmodule Glimesh.Schema.AccountTypes do
 
   import Absinthe.Resolution.Helpers
 
+  alias Glimesh.AccountFollows
   alias Glimesh.Avatar
   alias Glimesh.Repo
   alias Glimesh.Resolvers.AccountResolver
@@ -40,10 +41,10 @@ defmodule Glimesh.Schema.AccountTypes do
       config(fn args, _ ->
         case Map.get(args, :streamer_id) do
           nil ->
-            {:ok, topic: [Glimesh.AccountFollows.get_subscribe_topic(:follows)]}
+            {:ok, topic: [get_subscribe_topic(:follows)]}
 
           streamer_id ->
-            {:ok, topic: [Glimesh.AccountFollows.get_subscribe_topic(:follows, streamer_id)]}
+            {:ok, topic: [get_subscribe_topic(:follows, streamer_id)]}
         end
       end)
     end
@@ -81,14 +82,14 @@ defmodule Glimesh.Schema.AccountTypes do
 
     field :count_followers, :integer do
       resolve(fn user, _, _ ->
-        followers = AccountFollows.count_followers(user)
+        followers = count_followers(user)
         {:ok, followers}
       end)
     end
 
     field :count_following, :integer do
       resolve(fn user, _, _ ->
-        following = AccountFollows.count_following(user)
+        following = count_following(user)
         {:ok, following}
       end)
     end
--- a/lib/glimesh/graphql/schema/account_types.ex
+++ b/lib/glimesh/graphql/schema/account_types.ex
@@ -25,19 +25,11 @@ defmodule Glimesh.Schema.AccountTypes do
       arg(:username, :string)
       resolve(&AccountResolver.find_user/2)
     end
-
     @desc "List all follows or followers"
     field :followers, list_of(:follower) do
       arg(:streamer_username, :string)
       arg(:user_username, :string)
       resolve(&AccountResolver.all_followers/2)
-    end
-
-    @desc "Count all follows or followers"
-    field :followers, list_of(:follower) do
-      arg(:streamer_username, :string)
-      arg(:user_username, :string)
-      resolve(&AccountResolver.count_followers/2)
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/glimesh/graphql/schema/account_types.ex
```diff
@@ -32,6 +32,13 @@ defmodule Glimesh.Schema.AccountTypes do
       arg(:user_username, :string)
       resolve(&AccountResolver.all_followers/2)
     end
+
+    @desc "Count all follows or followers"
+    field :followers, list_of(:follower) do
+      arg(:streamer_username, :string)
+      arg(:user_username, :string)
+      resolve(&AccountResolver.count_followers/2)
+    end
```
clone1018, line 41: I think this approach would end up with us overwriting the existing followers() query. 

Instead it may be a good idea to add a countFollowers countFollowed to the existing User object? What are your thoughts?

clone1018 (changes requested):

mja00 (changes requested):
Only this one thing :D 

Once the alias is added to the `account_types.ex` file it should be all good to go!

For people looking at this in the future, the query from this PR to get following is(can replace myself with a user):
```
query {
  myself {
    following {
      streamer {
        username
      }
    }
  }
}
```
and followers is:
```
query {
  myself {
    followers {
      user {
        username
      }
    }
  }
}
```

lib/glimesh/graphql/schema/account_types.ex
```diff
@@ -80,6 +79,20 @@ defmodule Glimesh.Schema.AccountTypes do
       end)
     end
 
+    field :count_followers, :integer do
+      resolve(fn user, _, _ ->
+        followers = AccountFollows.count_followers(user)
```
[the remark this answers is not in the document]
AdamHebby: @mja00 Like I've just pushed? Have removed other AccountFollows references from that file too

lib/glimesh/graphql/schema/account_types.ex
```diff
@@ -41,10 +41,10 @@ defmodule Glimesh.Schema.AccountTypes do
       config(fn args, _ ->
         case Map.get(args, :streamer_id) do
           nil ->
-            {:ok, topic: [Glimesh.AccountFollows.get_subscribe_topic(:follows)]}
+            {:ok, topic: [get_subscribe_topic(:follows)]}
 
           streamer_id ->
-            {:ok, topic: [Glimesh.AccountFollows.get_subscribe_topic(:follows, streamer_id)]}
+            {:ok, topic: [get_subscribe_topic(:follows, streamer_id)]}
```
mja00, line 47:
```suggestion
            {:ok, topic: [AccountsFollows.get_subscribe_topic(:follows, streamer_id)]}
```

lib/glimesh/graphql/schema/account_types.ex
```diff
@@ -80,6 +80,20 @@ defmodule Glimesh.Schema.AccountTypes do
       end)
     end
 
+    field :count_followers, :integer do
+      resolve(fn user, _, _ ->
+        followers = count_followers(user)
```
mja00, line 85:
```suggestion
        followers = AccountFollows.count_followers(user)
```

lib/glimesh/graphql/schema/account_types.ex
```diff
@@ -80,6 +80,20 @@ defmodule Glimesh.Schema.AccountTypes do
...
+        {:ok, followers}
+      end)
+    end
+
+    field :count_following, :integer do
+      resolve(fn user, _, _ ->
+        following = count_following(user)
```
mja00, line 92:
```suggestion
        following = AccountFollows.count_following(user)
```

lib/glimesh/graphql/schema/account_types.ex
```diff
@@ -41,10 +41,10 @@ defmodule Glimesh.Schema.AccountTypes do
       config(fn args, _ ->
         case Map.get(args, :streamer_id) do
           nil ->
-            {:ok, topic: [Glimesh.AccountFollows.get_subscribe_topic(:follows)]}
+            {:ok, topic: [get_subscribe_topic(:follows)]}
```
mja00, line 44: Sorry, should've been more clear. Still getting used to doing proper reviews. When adding an alias it basically removes the necessity for the Glimesh. part, you'll still want everything after, so for this instance, it's 
```suggestion
            {:ok, topic: [AccountFollows.get_subscribe_topic(:follows)]}
```

mja00 (changes requested):
Sorry, this one's on me. Should be able to commit all of those suggestions on Github. I wasn't super clear with the review.

lib/glimesh/graphql/schema/account_types.ex
AdamHebby: @mja00 Is this a typo? Shouldn't `AccountsFollows` be `AccountFollows` - IE minus the 's'

mja00: Yep! Edited the comment and fixed.

mja00 (approved):
Approved. Will be merged pending tests. 👍🏻

PR #572, opened 2021-03-15
Fix invoking remote function within guard

Eluzn:
This fixes an issue where stripe.ex was unable to compile due to the create_subscription_invoice function.


clone1018:
The previous way of doing it is fine in Elixir 1.11.3, can you confirm?

Eluzn:
> The previous way of doing it is fine in Elixir 1.11.3, can you confirm?

Update from 1.10.3 -> 1.11.3 and seems to no longer through a compilation error.

Can be closed :)

PR #575, opened 2021-03-16
Add delete chat message button

### LINKED ISSUE #550: Delete chat message button
A button to delete a single message from a user

mja00:
Adds a delete chat message button to the chat. :)
Closes #550 

image

### DIFF
--- a/assets/js/hooks/Chat.js
+++ b/assets/js/hooks/Chat.js
@@ -115,6 +115,11 @@ export default {
             let offendingChatMessage = chatMessages.querySelectorAll(`[data-user-id=${CSS.escape(offendingUserID)}]`);
             // Have to hide them otherwise the tooltip gets stuck on removing the element
             offendingChatMessage.forEach(e => e.hidden = true);
+        });
+
+        this.handleEvent("remove_deleted_message", (e) => {
+            let offendingMessageID = e["message_id"];
+            document.getElementById(offendingMessageID).hidden = true;
         })
 
     }
--- a/assets/js/hooks/Chat.js
+++ b/assets/js/hooks/Chat.js
@@ -118,8 +118,7 @@ export default {
         });
 
         this.handleEvent("remove_deleted_message", (e) => {
-            let offendingMessageID = e["message_id"];
-            document.getElementById(offendingMessageID).hidden = true;
+            document.getElementById(e["message_id"]).hidden = true;
         })
 
     }
--- a/lib/glimesh/chat.ex
+++ b/lib/glimesh/chat.ex
@@ -159,11 +159,11 @@ defmodule Glimesh.Chat do
         %User{} = moderator,
         %Channel{} = channel,
         %User{} = message_author,
-        message_id
+        message
       ) do
     with :ok <- Bodyguard.permit(__MODULE__, :delete, moderator, channel) do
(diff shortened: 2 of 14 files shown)


lib/glimesh/chat.ex
```diff
@@ -388,4 +422,13 @@ defmodule Glimesh.Chat do
 
     {:ok, bad_user}
   end
+
+  defp broadcast_delete({:ok, channel_id, message_id}, :message_deleted) do
```
clone1018, line 426: Do we need `channel_id` here? Or can we broadcast the chat_message instead like the above `defp broadcast({:ok, chat_message}, event) do`

lib/glimesh/chat.ex
```diff
@@ -365,10 +388,21 @@ defmodule Glimesh.Chat do
     Repo.update_all(query, set: [is_visible: false])
   end
 
+  defp delete_chat_message(%Channel{} = channel, message_id) do
+    query =
+      from m in ChatMessage,
+        where:
+          m.is_visible == true and
+            m.channel_id == ^channel.id and
+            m.id == ^message_id
```
clone1018, line 397: I suggest refactoring this function to take in a %ChatMessage{} object alone, and then you can simplify it to:
```
chat_message
|> Repo.update(set: [is_visible: false])
```

a ChatMessage object has everything it needs to do this query, and is unique across all channels because of it's id.

assets/js/hooks/Chat.js
```diff
@@ -105,6 +105,11 @@ export default {
             let offendingChatMessage = chatMessages.querySelectorAll(`[data-user-id=${CSS.escape(offendingUserID)}]`);
             // Have to hide them otherwise the tooltip gets stuck on removing the element
             offendingChatMessage.forEach(e => e.hidden = true);
+        });
+
+        this.handleEvent("remove_deleted_message", (e) => {
+            let offendingMessageID = e["message_id"];
+            document.getElementById(offendingMessageID).hidden = true;
```
clone1018, line 112:
```suggestion
            document.getElementById(e["message_id"]).hidden = true;
```

Since we do not do anything else with offendingMessageID

test/glimesh/chat_test.exs
```diff
@@ -238,6 +239,15 @@ defmodule Glimesh.ChatTest do
       assert length(Chat.list_chat_messages(channel)) == 1
     end
 
+    test "delete deletes message", %{channel: channel, moderator: moderator, user: user} do
```
clone1018, line 242:
```suggestion
    test "delete_message/4 deletes message", %{channel: channel, moderator: moderator, user: user} do
```

clone1018: Since this is a context test

clone1018 (changes requested):
Thanks for the work @mja00 ! Couple of minor changes here to match some of our other code.

In addition, can you add the delete message function to the API so both authors can use them? Short timeout should be a straightforward example to follow, and I'd expect the input of the API to allow you to delete a message by it's ID since this just handles a single message.

Thanks!

mja00:
Tomorrow(3/28/21) is when I plan on fixing this up. :D

mja00:
@clone1018

clone1018 (approved):

PR #576, opened 2021-03-17
Subcategories

### LINKED ISSUE #570: Add Subcategories for descriptive filtering
We've been having some trouble getting tags to work exactly like we wanted. It seems most users are looking to describe two things: who they are, and what they are doing. Our current tag system throws both of these concepts together into one and means that users get confused on what tags to add. 

The proposal is to add basically a subcategory of tags that limit your selection of "what you are doing", like so:

image

Examples:

- Gaming - Game Title Tag: World of Warcraft, Final Fantasy XIV, Apex Legends
- Art - Style Tag: Digital, Pastels, Mixed Media
- Tech - Topic Tag: Programming, Video Editing
- Channel Tags - LGBTQIA+, Family Friendly, Leveling, Chill

This means adding a new option for selecting the subcategory of your stream, and then maintaining it as you move about activities. It's likely that this system will mean you are changing your subcategory frequently and your tags less frequently.

There's still some unanswered questions, but I'd like to explore them through code instead of discussion.

Questions:
* How do we ensure subcategories are tightly moderated / maintained? No spelling issues, etc
* Where do tags like: Horror, FPS, Multiplayer fit in? Are they available in the global namespace? 
* How does filtering show subcategories vs tags?

clone1018:
Fixes #570


clone1018:
@mja00 I apologize.

mja00 (approved):
Good to go

PR #578, opened 2021-03-17
Limit most recent 5 messages to the last 60 mintues

clone1018:


mja00 (approved):

haydenmc:
Hey friends,

I missed the rationale for this change. For places like chickencam, it's really valuable to see messages folks may have left a long time ago. Howcome we're getting rid of them?

clone1018:
We've had tons of feedback from the community that it doesn't make sense for most streams. It's actually been one of the most requested features so far. 

For chickencam, I suppose you could either overlay a popup chat, or use something like PixelChat to get the historical messages. All messages are still available from the API, this just makes users popping into streams less likely to be confused.

clone1018:
The other option is to keep chat messages around until the end of the stream, but this does add a performance implication because we have to look up the most recent stream, and when it ended.

Heiwa1580:
Is it plausible to have a load older message button? Not sure of the implications of a button that would just bring up older messages  at a time or something

clone1018:
It's possible of course, we just need to make a time / effort decision.

haydenmc:
> It's actually been one of the most requested features so far.

To clarify, I think what's actually been requested is a "clear chat" feature, not specifically "only show historical messages within the last 60 minutes." Though the latter might help address the former, it gives less control to the streamer.

Perhaps the chickencam scenario is the odd man out, but I do think this particular solution is a 'big hammer' workaround, and if we get a little more creative, we can solve for everyone. For example, adding a date separator between messages that were sent on a previous day (right now timestamps only show the time, not date, so it can be difficult to disambiguate old messages from new). Or actually adding the ability for streamers/moderators to 'clear chat,' causing clients to only show messages past the particular date/time when that button was last pressed. 

In addition to chickencam, I think there is a possibility that there are a number of low viewership streams out there that are motivated by seeing old chat, and could possibly be disappointed by it going away, and we wouldn't really see any indication of that until we take it away.

But also, I could be completely wrong. In which case I can be overruled. :)

clone1018:
I understand your point, I'm going to close this PR until we figure out how we can fit another solution in.

PR #581, opened 2021-03-19
Always create a new oauth token

### LINKED ISSUE #547: OAuth not conforming to spec; returns existing token instead of granting a new one
Per @danstiner's comment [here](https://github.com/Glimesh/janus-ftl-plugin/pull/78#issuecomment-787216934):

> [...] [glimesh.tv](https://github.com/Glimesh/glimesh.tv) doesn't quite follow the OAuth 2.0 spec for granting tokens. Which is understandable, I feel libraries do a bad job of explaining the flow. Basically Glimesh.tv will return your current access token (if it's still active) instead of granting a new one: https://github.com/Glimesh/glimesh.tv/blob/dcfff8d60743a603ca24ebb156e08287920c7d6a/lib/glimesh/oauth/oauth_token_handler.ex#L149
>
> The spec says every request should get a new access token, and the library https://github.com/danschultzer/ex_oauth2_provider used assumes this as well. So we end up in a weird place where you request one token, then request a new one and hour later and get exactly the same response with the same token and same `expires_in` value, despite the token being one hour closer to expiring: https://tools.ietf.org/html/rfc6749#section-5.1
>
> The [Janus-FTL] glimesh service connection works around this by using `created_at`, which works for now but has a couple problems. First, it's not part of the OAuth spec so other clients won't follow this approach, they'll expect a new token. Second, clients can't preemptively get a new access token before the current one expires to avoid interruptions (because they'll just get back the current token until it expires). Third, a fun finding I had was `created_at` is in the timezone of the Glimesh.tv server. Luckily that's UTC for the actual production service but I had some really fun issues locally when `multipass` created VMs in the PST timezone. Correctly issuing a new token every time and just relying on `expires_in` should avoid all these problems and work better with oauth client libraries that our users will probably be using to connect to glimesh.tv.

danstiner:
The [Oauth2 RFC ](https://tools.ietf.org/html/rfc6749#section-5.1) says every request to issue a token should get a new access token and the library being used here (https://github.com/danschultzer/ex_oauth2_provider) is designed around that assumption as well.

Because we return an existing token if there is one, we get crazy behavior like every single ingest node sharing the same access token (because they all send auth requests with the same client credentials). I don't have proof but I'd bet this causes strange behavior when the token is expiring. With this change ftl ingest servers will each get their own token, plus they will be able to request a new token before the current one expires and avoid any expiration races.

Fixes #547, see that issue for more context.

Might also help with https://github.com/Glimesh/janus-ftl-plugin/issues/112

Tested locally quick but probably could use some more validation, the main downside of this is the token table will grow a bit, and someone could potentially do a denial-of-service by creating a huge number of tokens.

### DIFF
--- a/lib/glimesh/oauth/oauth_token_handler.ex
+++ b/lib/glimesh/oauth/oauth_token_handler.ex
@@ -62,7 +62,7 @@ defmodule Glimesh.OauthHandler.Token do
 
     result =
       {:ok, %{client: client, resource_owner: resource_owner, scopes: post_request["scope"]}}
-      |> maybe_create_access_token(config)
+      |> create_access_token(config)
 
     case result do
       {:ok, {:error, error}} -> Error.add_error({:ok, params}, error)
@@ -132,9 +132,9 @@ defmodule Glimesh.OauthHandler.Token do
   defp validate_redirect_uri({:ok, params}, _config),
     do: Error.add_error({:ok, params}, Error.invalid_request())
 
-  defp maybe_create_access_token({:error, _} = error, _config), do: error
+  defp create_access_token({:error, _} = error, _config), do: error
 
-  defp maybe_create_access_token(
+  defp create_access_token(
          {:ok, %{resource_owner: resource_owner, client: application, scopes: scopes}},
          config
        ) do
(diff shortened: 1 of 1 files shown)


clone1018:
I could be doing something wrong here, but when I manually send multiple curl POST's for a new token on this branch it returns the same token. 

Params:
```
client_id=whatever,client_secret=whatever,grant_type=client_credentials,scope=public
```

danstiner:
I had tested it worked a long time ago with something similar, let me actually verify this branch lol

danstiner:
Yeah I'm just wrong, this is the wrong codepath to touch.

The client credential grants go straight to https://github.com/danschultzer/ex_oauth2_provider which internally checks for an existing token, so it's not easy to fix this.

Going to go ahead and close for now, but I would definitely support swapping out that library for a better one.

PR #582, opened 2021-03-24
Adding query caching into ETS for common unauthed requests

clone1018:


mja00 (approved):
Really nice thing to add

PR #592, opened 2021-03-28
Handling slightly more than zero packet loss before alerting user

clone1018:


lib/glimesh_web/live/user_live/stream.ex
```diff
@@ -104,13 +104,18 @@ defmodule GlimeshWeb.UserLive.Stream do
   end
 
   def handle_event("lost_packets", %{"uplink" => uplink, "lostPackets" => lostPackets}, socket) do
+    message =
+      if lostPackets > 3,
+        do:
+          gettext(
+            "We're detecting some networking problems between you and the streamer. You may experience video drops, jitter, or other issues!"
```
mja00, line 111: Possibly add a "This is not a bug." to the end of the message to cut down on people just posting the message in the bugs channel. Either way, good to go from me :) 

Dan might have some more input since he's got the big brain.

mja00 (approved):

clone1018:
A user on Twitter had an idea of what this could look like as well: https://twitter.com/cbtec_XBX/status/1376247554644525057

mja00:
I wish Twitter compression didn't destroy that video but I do like that too.

lib/glimesh_web/live/user_live/stream.ex
danstiner: No no no it's clone with the big 🧠 ideas.

danstiner: I'd suggest raising the threshold though to at least 10, or honestly even higher.

When I've seen this message it's bursts of 5+ packets and I have a decent network.

And eventually it'd be nice to measure lost over expected percent, packet inter arrival jitter, etc

danstiner (approved):

lib/glimesh_web/live/user_live/stream.ex
danstiner: Alternatively go with the super subtle route of changing the "signal" icon to one indicating no signal. Given this notification isn't really actionable to the viewer or streamer I don't think we need to be in their face with it.

lib/glimesh_web/live/user_live/stream.ex
```diff
@@ -104,13 +104,18 @@ defmodule GlimeshWeb.UserLive.Stream do
   end
 
   def handle_event("lost_packets", %{"uplink" => _uplink, "lostPackets" => lostPackets}, socket) do
+    message =
+      if lostPackets > 3,
```
clone1018, line 108:
```suggestion
      if lostPackets > 6,
```

lib/glimesh_web/live/user_live/stream.ex
clone1018: Updated the threshold to 6 instead, I do want to keep some user visible message around though until we introduce NACKs. The Chrome browser especially does not handle these dropped packets at all, and will unload the video client making for a confusing experience for the viewer.

PR #593, opened 2021-03-29
Markdown Preview

aMytho:
Adds a markdown preview and a view profile button. Unlike editing, the tab just gets longer instead of adding a scrollbar. I didn't test very wide images but the scaling seems to be fine. 

image

### DIFF
--- a/lib/glimesh_web/live/user_settings/components/profile_settings_live.html.leex
+++ b/lib/glimesh_web/live/user_settings/components/profile_settings_live.html.leex
@@ -150,8 +150,12 @@
 
                             <%= textarea f, :profile_content_md, [class: "form-control mb-4 text-monospace", id: "textarea_edit", rows: 20] %>
                         </div>
-                        <div class="tab-pane form-control mb-4" id="preview">
-                            <p><%= gettext("Coming soon!") %></p>
+                        <div class="tab-pane form-control mb-4" style="height:auto" id="preview">
+                            <%= if @user.profile_content_html do %>
+                                <%= raw(@user.profile_content_html) %>
+                            <% else %>
+                                <p>No Profile Content</p>
+                            <% end %>
                         </div>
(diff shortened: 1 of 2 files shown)


lib/glimesh_web/live/user_settings/components/profile_settings_live.html.leex
```diff
@@ -150,15 +150,20 @@
 
                             <%= textarea f, :profile_content_md, [class: "form-control mb-4 text-monospace", id: "textarea_edit", rows: 20] %>
                         </div>
-                        <div class="tab-pane form-control mb-4" id="preview">
-                            <p><%= gettext("Coming soon!") %></p>
+                        <div class="tab-pane form-control mb-4" style="height:auto" id="preview">
+                            <%= if @user.profile_content_html do %>
```
clone1018, line 154: From the looks of this, it is not a live preview but instead of a rendering of the user's previous profile content. Am I reading it right?

aMytho: Yeah, if you were to open a new tab to edit and preview it you wouldn't see the changes. I assumed that would be fine since the only place you can edit the profile was on that page, as long as you save the changes they will appear. I can try to make it appear without needing to save it.

clone1018: Yeah, the intention of the Preview tab is to be a live preview of your profile to make editing easier.

aMytho:
It shows up live now. If there is a better way to do this let me know. Currently it sends the markdown by `phx-change` but it is only converted when the user presses preview. I had to use `:edit_state` to switch between preview/edit. The tabs kept updating and resetting themselves so I removed them. The second input is hidden so if the user presses update settings the form can still get the markdown data (it was removed when preview was showing).

PR #595, opened 2021-03-29
Fix OAuth2 to not break whenever we add more scopes for existing applications

wolfcomp:
Continues #332 as I'm moving the branch over from my own fork to the main

- [x] Implement new OAuth flow
- [ ] Add in migrations for new OAuth flow
- [ ] Test both OAuth flow if they work

To Do
Remove old OAuth flow entirely this will be done at a later date, so 3rd party devs have time to change between the flows


wolfcomp:
Moved to #736

PR #619, opened 2021-04-15
Transactionally end streams

### LINKED ISSUE #610: Streams showing as offline even though they are live
Saw another instance of this tonight, wanted to document it a bit before I went to bed.

The channel's state was offline, with a null stream ID. However the streams table had a recent stream with a null ended_at, implying the channel was still online. This also allows Janus to continue to update us about metadata since the stream is still valid

clone1018:
Potentially fixes #610, but we wont know for sure since we haven't been able to reproduce it exactly...﻿

### DIFF
--- a/lib/glimesh/streams/channel.ex
+++ b/lib/glimesh/streams/channel.ex
@@ -53,14 +53,14 @@ defmodule Glimesh.Streams.Channel do
   def start_changeset(channel, attrs \\ %{}) do
     channel
     |> changeset(attrs)
-    |> put_change(:status, "live")
+    |> force_change(:status, "live")
   end
 
   def stop_changeset(channel, attrs \\ %{}) do
     channel
     |> changeset(attrs)
-    |> put_change(:stream_id, nil)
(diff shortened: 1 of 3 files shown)


lib/glimesh/streams/channel.ex
```diff
@@ -53,14 +53,14 @@ defmodule Glimesh.Streams.Channel do
   def start_changeset(channel, attrs \\ %{}) do
     channel
     |> changeset(attrs)
-    |> put_change(:status, "live")
+    |> force_change(:status, "live")
```
clone1018, line 56: The theory with the force_change is that we might be mutating a bad changeset. This would force the SQL to update even if the changeset is wrong.

mja00 (approved):
I had the bug during my lotro stream yesterday when my network dropped. My live tag was still there when my net came back but before OBS reconnected, then it went away once I reconnected. So it's definitely a thing of getting a stale changeset. Forcing should fix. :D

PR #625, opened 2021-04-16
Create channel button

EmberShadow:
Checks if 

a) The profile does not channel attached

b) If it is the logged in user is the owner of the profile

If these conditions are met - a create channel button displays on their profile.


mja00 (approved):
For anyone curious of what it looks like, when viewing your own profile: 
image
When not: 
image

PR #643, opened 2021-04-20
Popped out & mobile chat fixes

mja00:
Two fixes/things in this PR. 

Popped out chat now completely ignores the device width, so no longer will it format like it's on mobile(should help when docking chat in OBS).

The second thing is the colon in the chat bubble for mobile has been moved to after the timestamp to help with reading flow. 

Before:
image

After: 
image

Desktop in stream(chat next to stream) chat has not been affected by any of these changes. Popout just now behaves like it did before #591 was merged.

### DIFF
--- a/lib/glimesh_web/live/chat_live/index.ex
+++ b/lib/glimesh_web/live/chat_live/index.ex
@@ -59,7 +59,7 @@ defmodule GlimeshWeb.ChatLive.Index do
         |> assign(:show_timestamps, user_preferences.show_timestamps)
         |> assign(:show_mod_icons, user_preferences.show_mod_icons)
         |> assign(:user_preferences, user_preferences)
-        |> assign(:popped_out, session["popped_out"])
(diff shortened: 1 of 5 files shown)


lib/glimesh_web/live/chat_live/index.ex
```diff
@@ -59,6 +59,7 @@ defmodule GlimeshWeb.ChatLive.Index do
         |> assign(:show_timestamps, user_preferences.show_timestamps)
         |> assign(:show_mod_icons, user_preferences.show_mod_icons)
         |> assign(:user_preferences, user_preferences)
+        |> assign(:popped_out, session["popped_out"])
```
clone1018, line 62: If you give this a default, we can remove a lot of the code changed in the test file and make this component easier to use:
```suggestion
        |> assign(:popped_out, Map.get(session, "popped_out", false))
```

clone1018 (changes requested):

mja00:
@clone1018 this is ready for review again :)

PR #657, opened 2021-04-24
Cleaning up the new API

clone1018:
Refactoring new api to /api/graph
Adding complete coverage of all resolvers
Removing the old API routes from coverage
Cleaning up deprecated fields
Locks away new API until we're ready


wolfcomp (approved):
Cant see anything that should be wrong and can totally see why `isMod` would be disabled until something more performant is found.

PR #662, opened 2021-04-26
Add chat message validation

clone1018:
We get a fatal error whenever createChatMessage is called with a message longer than 255 characters (since that's our DB limit).

This PR shows ecto that limit, and tells Absinthe how to translate it into an error message. Ideally we should just drop in something like: https://hex.pm/packages/absinthe_error_payload in the future.

### DIFF
--- a/lib/glimesh/api.ex
+++ b/lib/glimesh/api.ex
@@ -8,4 +8,15 @@ defmodule Glimesh.Api do
   def error_not_found, do: {:error, @error_not_found}
 
   def error_access_denied, do: {:error, @error_access_denied}
+
+  @doc """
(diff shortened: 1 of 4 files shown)


mja00 (approved):

Eein:
Just a personal note about the package - I've found that Kronky and `absinthe_error_payload` are fairly clunky in practice and I've had issues upgrading ecto because of them. Writing a simple module (like you have) that traverses errors and extracting it to a common module is the more testable and predictable - 

I've used the exact same code you wrote professionally (from wherever it came from) with great success in absinthe for a couple of years now :)

PR #665, opened 2021-04-28
Chat message metadata

Eein:
A couple of important notes off the bat:
- upgrades waffle_ecto from `0.0.9` to `0.0.10`, which upgrades some child dependencies, check the mix lock (this was the issue i was having with  `Glimesh.ChannelPoster.Type` `:poster` code reloading
- adds a new `:metadata` jsonb field to chatmessage
- adds faker to `:test`
- adds `ex_machina` to the project and some initial factories to support

The object of this PR is to take some of existing behavior around chat messages and loading badges and move the logic into metadata so it can be loaded without querying on read. This should save 2 or three calls per message load, but doesn't solve the subscription issue (yet)

I was mindful about retaining fallbacks for existing functionality for messages without metadata - but in the future there could be removals here to add a fraction of performance on rerenders

I've also extracted the badge renders into their own modules so they can be used a little more cleanly (and also for reuse, test or refactor)

There's plenty of additional refactoring opportunities here but i'm looking for some opinions on this draft before I go too far.


Eein:
This PR does quite a lot - I wouldn't feel bad if it was stomped out until additional thought was put in the the design/architecture of the channel page. I feel many of the issues that this solves could be additionally solved by rethinking the live views & data flow in general, but it would likely be a big undertaking.

However this should work just fine if a minor improvement is acceptable. 

I really don't think this PR should get any bigger because of the sheer amount of dependencies added (for personal reasons) and the database migration required to support the metadata.

Fixing the subscription issue would be viable down the road if this PR is in place, since the subscription would query on new chat message creation, rather than on every `GET`

lib/glimesh/chat.ex
```diff
@@ -34,15 +34,23 @@ defmodule Glimesh.Chat do
   def create_chat_message(%User{} = user, %Channel{} = channel, attrs \\ %{}) do
     with :ok <- Bodyguard.permit(__MODULE__, :create_chat_message, user, channel) do
       if allow_link_in_message(channel, attrs) do
+        platform_subscriber = Glimesh.Payments.is_platform_subscriber?(user)
```
clone1018, line 37: I think this may be the wrong field. Platform subscriber is for Glimesh.tv directly, and is not specific to any one chat message. Channel subscriber is for the individual channel, and the users get a Sub icon when they have an active subscription.

lib/glimesh/chat/effects.ex
```diff
@@ -101,31 +110,42 @@ defmodule Glimesh.Chat.Effects do
...
+
+  def render_channel_badge(%Message{metadata: %Metadata{subscriber: true}}) do
+    ChannelSubscriberBadge.render()
+  end
+
+  def render_channel_badge(%Message{metadata: %Metadata{admin: true}}), do: ""
+  def render_channel_badge(%Message{metadata: nil}), do: nil
+  def render_channel_badge(%Message{metadata: %Metadata{} = _metadata}), do: nil
+
   def render_channel_badge(channel, user) do
```
clone1018, line 136: Why do we still keep this function around? Shouldn't the above take care of all of the cases?

priv/repo/migrations/20210428020716_add_metadata_to_chat_messages.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Glimesh.Repo.Migrations.AddMetadataToChatMessages do
+  use Ecto.Migration
+
+  def change do
+    alter table(:chat_messages) do
+      add :metadata, :map
+    end
```
clone1018, line 7: Should this migration all update all of the existing chat messages?

clone1018 (changes requested):
Looks really good! The testing improvements are very nice.

lib/glimesh/chat/effects.ex
Eein: Assuming I'm not writing a migration here, it was to handle falling back to the old behaviour if required.
Since your preference is to write a migration --- can definitely remove this.

lib/glimesh/chat.ex
Eein: Sounds good! I'll remove this and retest any further assumptions i had here. 
This looks like a good candidate for reading this data in the user session, rather than the chat message function

priv/repo/migrations/20210428020716_add_metadata_to_chat_messages.exs
Eein: Can do, and similar response to https://github.com/Glimesh/glimesh.tv/pull/666#discussion_r631881980

clone1018:
Merged in https://github.com/Glimesh/glimesh.tv/pull/796

PR #666, opened 2021-04-28
Social account username sanitization

### LINKED ISSUE #48: socials on profile page should automatically strip out @ symbols
when adding your social accounts it does not strip out the @ character at the beginning of the account name. it is quite common for users to refer to their twitter/instagram page(maybe others too) prefixed with an @ symbol. when the link if generated for the profile page it also includes the extra character

Eein:
- adds sanitiziation to public facing social buttons
- attempts to sanitize when connecting manually input social

This covers the following cases
- "@username" -> "username"
- "@@@@@username" -> "username"
- "user@name@" -> "user@name@"
- "////username////" -> "username"
- "    username    " -> "username"

closes #48

Edit:
Adding guilded.gg sanitize under a :guilded function param for perf
See the following regex:
https://rubular.com/r/KwEKgHNxi3y3ej

What this does not fix:
- A frontend liveview form validation that lets the user know to only add the username part.
- The value in the form field when editing the guilded "url" - if they put `https://guilded.gg/TheKeep` then it will still show like that when editing preferences until saved again.

### DIFF
--- a/lib/glimesh_web/live/user_live/components/social_buttons.ex
+++ b/lib/glimesh_web/live/user_live/components/social_buttons.ex
@@ -1,13 +1,15 @@
 defmodule GlimeshWeb.UserLive.Components.SocialButtons do
   use GlimeshWeb, :live_view
 
+  alias Glimesh.Socials.Sanitizer
+
   @impl true
   def render(assigns) do
     ~L"""
     <%= if @twitter_social  do %>
     <li rel="ugc" class="list-inline-item" data-toggle="tooltip"
         title="<%= gettext("Linked Twitter Account @%{username}", username: @twitter_social.username) %>">
-        <a href="https://twitter.com/<%= @twitter_social.username %>" target="_blank">
+        <a href="https://twitter.com/<%= Sanitizer.sanitize(@twitter_social.username) %>" target="_blank">
             <span class="fa-stack">
                 <i class="fas fa-certificate fa-stack-2x"></i>
                 <i class="fab fa-twitter fa-stack-1x" style="color:white"></i>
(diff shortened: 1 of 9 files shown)


clone1018:
Looks good, the main thought I had is should we be doing this sanitization before we save to the DB so we can use the same values on the UI, DB, and API?

Eein:
Sure, the current code shouldn't be prone to false positives - 

This is mostly just a monkey patch for the meantime but im happy to take on the rest of the feature

Eein:
This one should be done :) enjoy

lib/glimesh_web/live/user_live/components/social_buttons.ex
```diff
@@ -1,13 +1,15 @@
...
+  alias Glimesh.Socials.Sanitizer
+
   @impl true
   def render(assigns) do
     ~L"""
     <%= if @twitter_social  do %>
     <li rel="ugc" class="list-inline-item" data-toggle="tooltip"
         title="<%= gettext("Linked Twitter Account @%{username}", username: @twitter_social.username) %>">
-        <a href="https://twitter.com/<%= @twitter_social.username %>" target="_blank">
+        <a href="https://twitter.com/<%= Sanitizer.sanitize(@twitter_social.username) %>" target="_blank">
```
clone1018, line 12: Isn't this essentially sanitizing it twice? Shouldn't we migrate all the existing records to be safe?

clone1018 (changes requested):
Looks great, one question though.

lib/glimesh_web/live/user_live/components/social_buttons.ex
Eein: Thats correct, this is just a monkey patch to correct the records that exist. 

I can write a data migration if preferred; do you prefer running these types of non-schema migrations on an iex in prod or making a db migration? (Or any other way? I know people usually have a preference)

PR #668, opened 2021-04-29
Fixes a bug in Janus Debug pane

Eein:
The incorrect parameter was being sent over, causing something to crash for some reason, since stream should pretty much always exist, this should no longer happen.

I chose not to include moving to live components for the debug window until i trial some concepts - since the existing page setup is not exactly what i'm used to in live view.

currently all of the state is handled at the top, but what I think should likely happen is that each stateful component should handle its own loading of data. Going to trial this on a boilerplate project to see if i can get it working the way i want.


clone1018 (approved):
Looks good, thanks!

PR #680, opened 2021-05-12
Maint/speed up GitHub actions

Eein:
Caching build dependencies makes new builds more efficient on test in actions as it only recompiles things it needs to.
This should improve times once a build is available by around ~2-3m.

I'd also recommend enabling workflow dispatch on your releases action (if you can gate it to only admins) as it allows you to roll back deploys to previous revisions quite easily. It's also very helpful for testing actions - but i'll leave it up to you.

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -1,6 +1,6 @@
 name: Elixir CI
 
-env: 
+env:
   MIX_ENV: test
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


Eein:
image
This is the best case, roughly 2.5m~

There's some improvements to be made by extracting some additional things like code quality out into separate actions but i'll leave those for another day.

clone1018:
Awesome stuff!

PR #684, opened 2021-05-16
Update elixir.yml

Eein:


.github/workflows/elixir.yml
```diff
@@ -50,16 +50,16 @@ jobs:
         key: ${{ runner.os }}-build-${{ hashFiles('**/mix.lock') }}
         restore-keys: ${{ runner.os }}-build-
 
-    - name: Compile Dependencies
-      if: steps.build-cache.outputs.cache-hit != 'true'
-      run: MIX_ENV=test mix compile
-
     - name: Install dependencies
       run: |
```
wolfcomp, line 54: Maby add `if: steps.build-cache.outputs.cache-hit != 'true'` here aswell?

Issue #637, opened 2021-04-19
Custom Channel Emotes


clone1018:
Since we've launched the cute little glimdrops in the bottom of the chat, people have demanded their own custom emotes to make their channel feel like home. Soon you will be able to bring your personality to the platform with your own channel emotes. These channel emotes can be anything as long as it doesn’t break any of our ToS or RoC. We are discussing the exact implementation to figure out the best approach to give streamers the flexibility they need in their channel.

_This feature is still being groomed and could change frequently._

clone1018:
Initial global emote work is happening in https://github.com/Glimesh/glimesh.tv/issues/637

Next couple of proposed phases:
1. Deploy new global emote system
2. Add emotes to the API
3. Add channel-specific, channel-only emotes
4. Allow channel emotes to be used globally in certain situations

clone1018:
A majority of these features are on production.

PR #697, opened 2021-05-30
Custom Channel Emotes

clone1018:
Very first version of custom channel emotes. Lays the groundwork of virtually all of the required pieces for emotes, but doesn't implement all of them.

For now the basic implementation allows for:
* Channels may have a single static emote. 
* Any uploaded emotes are sent to a GCT dashboard for review.
* You can only use the channel emote inside the channels chat.
* The emote is available to anyone, not just subscribers. 

Known Problems
* Validation is wonky, error messages duplicate and are not always helpful. 

As far as SVG security goes, I believe it's mostly mitigated by most browsers. However since we cannot manually set CSP headers with most CDNs, we're utilizing both sanitizing the image with svgo, serving it through an img tag so js cannot run, and setting the content-disposition to attachment so the browser wont automatically run it.


clone1018:
As mentioned in https://github.com/Glimesh/glimesh.tv/issues/637 -- this work is 2/4 of the PR's we'll need for "full" channel emotes.

mja00 (approved):

PR #710, opened 2021-06-08
Start of hosting

mja00:


mja00:
I do want to start work on this again, however, I believe I will want to wait for #724 as those redesign changes would make it somewhat easier to hot-swap out elements for hosting properly. :D

clone1018:
Merged in https://github.com/Glimesh/glimesh.tv/pull/786

Issue #720, opened 2021-06-28
Add #streaming-help to the packet loss modal


clone1018:
Re: https://discord.com/channels/727111257953402891/727484488266940488/859207591518273566

mja00:
Added by #728

PR #728, opened 2021-07-05
Add discord link on packet loss message

aMytho:
PR adds a link to the discord if there is packet loss.

I don't know if this is the best way to do it, but Phoenix.html.raw was the only way that I found to add an html tag. If there is a better way to do it let me know and I'll change it. 

Adds #720

### DIFF
--- a/lib/glimesh_web/live/user_live/stream.ex
+++ b/lib/glimesh_web/live/user_live/stream.ex
@@ -111,15 +111,15 @@ defmodule GlimeshWeb.UserLive.Stream do
     message =
       if lost_packets > 6,
         do:
-          gettext(
-            "We're detecting some networking problems between you and the streamer. You may experience video drops, jitter, or other issues!"
-          ),
+        gettext(
+          "We're detecting some networking problems between you and the streamer. You may experience video drops, jitter, or other issues! If this continues, the streamer is recommended to submit a ticket in #streaming-help in our <a href=\"https://discord.gg/Glimesh\">Discord</a> server."
+        ),
         else: nil
 
     {:noreply,
      socket
      |> update(:lost_packets, &(&1 + lost_packets))
-     |> assign(:player_error, message)}
+     |> assign(:player_error, Phoenix.HTML.raw(message))}
   end
 
   def handle_event("lost_packets", _, socket) do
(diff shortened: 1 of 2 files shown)


lib/glimesh_web/live/user_live/stream.ex
```diff
@@ -111,9 +111,10 @@ defmodule GlimeshWeb.UserLive.Stream do
     message =
       if lost_packets > 6,
         do:
-          gettext(
-            "We're detecting some networking problems between you and the streamer. You may experience video drops, jitter, or other issues!"
-          ),
+        gettext(
+          "We're detecting some networking problems between you and the streamer. You may experience video drops, jitter, or other issues! If this continues, the streamer is recommended to submit a ticket in #streaming-help in our"
+        ) <> " <a href=\"https://discord.gg/Glimesh\">Discord</a>."
+        |> Phoenix.HTML.raw(),
```
clone1018, line 117: Thanks for getting this done Mytho!

Since `gettext()` can be "user input" via our translation service, we don't want to call raw() with it or we could be injected. What I normally do is use UI patterns that split the link & the text. What we may want to do here is add an independent button for _all_ warnings we show inside the debug modal here: https://github.com/Glimesh/glimesh.tv/blob/dev/lib/glimesh_web/live/user_live/stream.html.leex#L105-L107

Something like:
> We're detecting some networking problems between you and the streamer. You may experience video drops, jitter, or other issues! If this continues, the streamer is recommended to submit a ticket in #streaming-help in our Discord
> [Join Our Discord!] (this would be a button)

clone1018 (changes requested):

lib/glimesh_web/live/user_live/stream.ex
aMytho: Ok, do you want it below the alert class or below?
image


image

clone1018: The latter looks good to me. What do you think?

aMytho: I think either one is fine, but I'll go with the last one.

clone1018 (approved):

PR #729, opened 2021-07-07
Streamloots integration

clone1018:
Adds a Streamloots addon to channels to allow them to show multiple ways of supporting their stream. Also changes the name of the Subscribe button to Support. 

## Video Demo

https://user-images.githubusercontent.com/226638/124822856-7a6f1a00-df3e-11eb-8525-f418d609f003.mp4

## Screenshots

image
Introduces a new page called Addons under your channel. In the medium-term feature we'll move this to our Marketplace instead. 

Once you've set it up your viewers will be able to see the Streamloots option under the Support button:
image


lib/glimesh_web/live/channel_settings_live/addons.html.leex
```diff
@@ -0,0 +1,33 @@
+<div class="card">
```
clone1018, line 1: Translate this file

lib/glimesh_web/live/support_modal/support_modal.ex
```diff
@@ -0,0 +1,60 @@
...
+  alias Glimesh.Accounts
+
+  @impl true
+  def mount(_params, %{"streamer" => streamer, "user" => nil} = session, socket) do
+    if session["locale"], do: Gettext.put_locale(session["locale"])
+
+    channel = Glimesh.ChannelLookups.get_channel_for_user(streamer)
+    can_receive_payments = Accounts.can_receive_payments?(streamer)
+
+    {:ok,
```
clone1018, line 13: Should be able to combine these two function's assigns since they are so similar.

clone1018:
Going ahead with this, we can refactor later :)

PR #730, opened 2021-07-08
Adding lazy loading to followers and following page

### LINKED ISSUE #726: Following / Follower routes should paginate / lazyload
From looking at PromEx metrics, I've seen some spikes in Phoenix HTTP request times.

Both of these routes are sized based on the count of users or followers:
* `/:username/profile/followers`
* `/:username/profile/following`

For users with a high number they are following (eg. 480+) I've seen request times taking ~7 seconds. This is reflected in the user experience as well with a high TTFB. It's similar when looking at your followers (~360) with a slow page load.

This further causes 100s of avatar requests all at once which isn't ideal.

It seems these pages should behave more like the user list page where the listing is lazy loaded as your scroll. This would improve the load performance I expect?

image

mja00:
Adds lazy loading to the followers and following page. Fixes #726 

For some reason my VSCode reset its line endings and code_quality got real mad so it decided to fix everything that VSC touched. Idk why but yeah. That's why 8 million files are touched.


clone1018 (approved):
I like this better at 6 files :D

PR #734, opened 2021-07-17
Prototype Stream page with Sticky Chat

Ramireztime:
Just another attempt at a possible layout of stream page. Design/layout departs from existing production site. As an initial commit, there are some issues with mobile layout (responsiveness).

image

[Video of earlier prototype, action similar to this PR]
https://cdn.discordapp.com/attachments/731685060012539924/865777718860054528/2021-07-16_19-06-16.mp4


Cykotiq:
title-and-stream-info-on-bottom should also be broken out into a separate commit

Ramireztime:
Alright, not entirely sure how to proceed with having this be a toggleable feature. I do have a version that uses CSS (avoids structure changes to HTML) but it's not at parity with this PR. I have some understanding of what needs to take place but not sure what that code looks like.

One major issue right now is that the chat is placed at the bottom of the page, due to changes to structure of HTML. A fix for this could be just having the chat component twice on the page and use Bootstrap utility classes to show/hide depending on screen size. I've gotten this to work, however only the first chat component that is loaded in code structure actually functions as intended. Not sure if this is something related to backend Elixir code.

As for the sidebar, the only changes to get that feature to revert back to old functionality would be to remove "position: sticky" (well, set it to relative, and remove top property).

clone1018:
> One major issue right now is that the chat is placed at the bottom of the page, due to changes to structure of HTML. A fix for this could be just having the chat component twice on the page and use Bootstrap utility classes to show/hide depending on screen size. I've gotten this to work, however only the first chat component that is loaded in code structure actually functions as intended. Not sure if this is something related to backend Elixir code.

We can swap out the HTML easily with Elixir, which would fix what it hooks the JS events to, however the problem would then be for any users who switch between maybe portrait and landscape mode on a tablet, where they would get bottom chat _and_ side chat. 

Honestly I'm not 100% convinced this _must_ be a feature flag, especially if we can get mobile sizing working and if we can figure out how to deal with screen size changes that a user would go through. Originally the feature flag was important if we were moving the stream titles & buttons around too.

clone1018:
Here's the diff if we decide to use the feature flag still: https://github.com/Glimesh/glimesh.tv/compare/4eb7851e98ee361ff245a80040e1d7345718a74c...protoype/example-sticky-chat-with-toggle

Ramireztime:
> Honestly I'm not 100% convinced this _must_ be a feature flag, especially if we can get mobile sizing working and if we can figure out how to deal with screen size changes that a user would go through. Originally the feature flag was important if we were moving the stream titles & buttons around too.

Initially I was thinking of changing placement, however Cykotiq made a good argument for keeping the existing layout, albeit with slight visual changes. It's still something we can test if the toggle feature makes it to production site. We can test a "ver 1" layout for a couple days/weeks, collect feedback from community, then ship a "ver 2", repeat feedback collection phase. Final layout would ultimately be determined by key people (and community feedback).

Ramireztime:
Approaching this, I knew that there would be some bulk added to overall code for stream page. The toggle feature works as intended, however there's still the lingering issues with mobile/tablet layouts as mentioned above for new layout.

- on mobile, chat shows up well below video (and channel footer) due to placement in code structure.
- tablet, there's still some issues that arise from a user switching between landscape and portrait.

PR #736, opened 2021-07-29
Switch to Boruta for Oauth 2.0 

clone1018:
Started with the work @wolfcomp created in https://github.com/Glimesh/glimesh.tv/pull/595 and then tied it together with the rest of the app.

Implements the proposal outlined in https://build.glimesh.tv/t/the-plan-for-the-great-oauth-migration/45 but will need to look at this throughout the rest of the week to make sure it all makes sense.

Items Remaining:
- [x] Client Scopes? 
- [x] More testing
- [x] More tests?


wolfcomp:
Client in Boruta does not have any sense of knowledge when it comes to scopes it is only on actual tokens it knows so in that way it would use a client credentials for this

wolfcomp:
Meaning if you connect with a client id you would only get basic read access and nothing more

clone1018:
> Meaning if you connect with a client id you would only get basic read access and nothing more

Yeah, that should still be the way it is for the socket API. We don't allow Client ID only with the actual GraphQL API. 

> Client in Boruta does not have any sense of knowledge when it comes to scopes it is only on actual tokens it knows so in that way it would use a client credentials for this

I'm speaking of this table: https://hexdocs.pm/boruta/Boruta.Oauth.Clients.html#c:authorized_scopes/1 -- Seems like you can set default requested scopes? I did not re-implement that however yet.

wolfcomp:
Yeah i see no need to have force for what scopes you are requesting really other than maybe public but that is about it

wolfcomp (commented):
I'm probably going to write another pr after this to include some more changes to the OAuth flow, but only for the authorization request flow and not the actual API OAuth flow where you can select what scopes they are requesting that they actually get access too.
This would only be used when authorizing a token to be issued to an application, and would make application users more aware of what scopes they get back to allow for features they make in their application if someone does not enable a scope to be used that is required for that feature in their application.
If you want an example for what I'm thinking of, you can look at Discords bot endpoint when you add bots to a server.

wolfcomp (approved):
Seems good cant see anything that should be changed

clone1018:
> Seems good cant see anything that should be changed

Thanks for checking it out!

PR #743, opened 2021-08-23
Removing 420

Heiwa1580:
We are removing 420 from the list as marijuana is not breaking TOS as long as it is legal in the users location. Therefore the term 420 was removed as it has not TOS or ROC violations in it


mja00:
Seems good to me as it's not violating anything.

Issue #719, opened 2021-06-28
Add homepage channels API endpoint


clone1018:
Add a list of channels that are currently on the homepage.

PR #748, opened 2021-08-30
Current Homepage Channels Query

ponkey364:
(fyi, this is the first bit of elixr I've ever written, although it's mostly a copy/paste job from other queries)

Done off the back of #719, with the aim of going in and implementing the homepage streams list in the mobile app.

For the most part, I looked at how the other queries work, especially the categories query, so I think it *should* be in line with what's expected here.

I just used a `[Channel]` rather than a connection because I didn't see the point in setting it up for 6 streams

### DIFF
--- a/lib/glimesh/api/resolvers/channel_resolver.ex
+++ b/lib/glimesh/api/resolvers/channel_resolver.ex
@@ -6,6 +6,7 @@ defmodule Glimesh.Api.ChannelResolver do
   alias Glimesh.ChannelCategories
   alias Glimesh.ChannelLookups
   alias Glimesh.Chat.ChatMessage
+  alias Glimesh.Homepage
   alias Glimesh.Streams
 
   @error_not_found "Could not find resource"
@@ -86,6 +87,10 @@ defmodule Glimesh.Api.ChannelResolver do
   end
 
   def find_channel(_, _), do: {:error, @error_not_found}
+
+  def list_homepage_streams(_, _) do
+    {:ok, Homepage.get_homepage()}
+  end
 
   # Streams
   def start_stream(_parent, %{channel_id: channel_id}, %{context: %{access: access}}) do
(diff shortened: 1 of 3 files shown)


clone1018:
Thanks for the PR! Looks good so far. 

> I just used a `[Channel]` rather than a connection because I didn't see the point in setting it up for 6 streams

Let's use a connection anyway, as we continue to grow out the homepage we don't want to introduce breaking changes to the API just to add more streams :P. This is what got us into trouble with the first API.

ponkey364:
I think that this attempt should be better, it has a connection now

lib/glimesh/api/schema/channel_types.ex
```diff
@@ -61,6 +61,11 @@ defmodule Glimesh.Api.ChannelTypes do
 
       resolve(&ChannelResolver.all_channels/2)
     end
+
+    @desc "List the channels currently on the homepage"
+    connection field :homepage, node_type: :channel do
```
clone1018, line 66:
```suggestion
    connection field :homepage_channels, node_type: :channel do
```

lib/glimesh/api/resolvers/channel_resolver.ex
```diff
@@ -87,6 +88,15 @@ defmodule Glimesh.Api.ChannelResolver do
 
   def find_channel(_, _), do: {:error, @error_not_found}
 
+  def list_homepage_streams(args, _) do
```
clone1018, line 91:
```suggestion
  def list_homepage_channels(args, _) do
```

lib/glimesh/api/schema/channel_types.ex
```diff
@@ -61,6 +61,11 @@ defmodule Glimesh.Api.ChannelTypes do
...
+      resolve(&ChannelResolver.list_homepage_streams/2)
```
clone1018, line 67:
```suggestion
      resolve(&ChannelResolver.list_homepage_channels/2)
```

clone1018 (changes requested):
Tested it locally, works great! Ready to merge after a couple of nitpicks. Thanks again!

PR #755, opened 2021-09-08
feature/pronouns

Ribenaberri:
Have added a new feature to the user settings page - pronouns. Users can select from a dropdown list their preferred pronouns.

None is default.  

If a pronoun other than "none" is selected this will also show as a badge under their user profile picture on their profile page. 

pronouns2

pronouns1


mja00:
Only requested a review from you clone since it's a first-time contributor and wanna make sure it's good for you. 

For me, it looks all good, pulled it locally and tested it. The only thing I could see is using the string `None` as the default value instead of just using `nil` as the migration will have to update every user to fill the slot but that might just be me nitpicking 😂

Ribenaberri:
unknown
unknown (3)

have added the capacity to display pronouns in the channel alongside language and mature tags. 

added modular choices for where pronouns can be displayed, the idea here is that if more places are added i.e. in chat containers, people can choose where their pronouns will appear giving full control.

mja00 (approved):

lib/glimesh/accounts/user.ex
```diff
@@ -10,6 +10,9 @@ defmodule Glimesh.Accounts.User do
   schema "users" do
     field :username, :string
     field :displayname, :string
+    field :pronoun_id, :string
+    field :pronoun_stream, :boolean
+    field :pronoun_profile, :boolean
```
clone1018, line 15: For these fields, can we give them a little more clarity? I'm thinking we change `pronoun_id` to `pronoun`, and append `show_` to the `pronoun_stream` and `pronoun_profile`. This way we don't think the `pronoun` field is an integer ID, and we know the stream & profile ones are visual toggles.

clone1018 (changes requested):
Looks good, this is the exact way I would have implemented it. Only one wordage change before we can merge it in!

lib/glimesh/accounts/user.ex
Ribenaberri: Changes made as requested. I also made the 2 exs files into one file, rather than across 2 as it was, made sense to have the pronouns stuff together in that way.

PR #767, opened 2021-10-15
Updates API Documentation

AdamHebby:
I'm generating a PHP Glimesh Client Library, using the API specification and pulling in everything, including descriptions and I noticed only about 1/4 of the fields have descriptions, so I aimed to fix that here.

Please can you check that my descriptions are accurate, and I haven't made any major boo-boo's


clone1018:
Sorry for the delay, I had a review ready but forgot to hit the button!

PR #775, opened 2021-11-20
Adding proper redirect on login

mja00:
Uses [navigation_history](https://hex.pm/packages/navigation_history) to allow us to jump back in the user's URL history. Using this we are able to redirect back to the user's previous page on login/signup/sign out. Tested with logging in, logging out, and signing up, all of them redirect to the page the user was previously on.

Also tested with trying to authorize an app on an account that isn't signed it, once the user logins in it redirects back to the authorization page without any issue.


mja00:
This test is only failing on Github, I think it's taking too long to respond

PR #776, opened 2021-11-24
Fixing banned users from showing on the profile lists

mja00:
Currently banned users will show on the `profile/following` and `profile/follower` pages due to a forgotten `where` statement. This PR adds that statement.


mja00:
Will redo a PR properly, instead of being lazy on github.

PR #790, opened 2022-01-04
Changing the stream title should also change the page title

### LINKED ISSUE #789: Changing the stream title from the channel page does not change the page title until refreshed.
When you change your stream title, the page title will still reflect the old stream title until you refresh the page.  We should update the page title when the new title is saved.

title-bug

MemoryLeakDeath:
Fixes #789  When changing the stream title the page title should now update to reflect the change.


clone1018:
Thanks for the bug fix!

PR #804, opened 2022-01-26
Add mutations to create/modify/delete follows

### LINKED ISSUE #718: Add Follow / Unfollow mutations to the API
We need to allow the currently authenticated user to follow / unfollow channels via the API.

danstiner:
Fixes #718, which will be useful both for the app I'm working on and for the official app.

Change Summary:
- Add `follow` scope to allow modifying follows for the current user
  - No current scope covered this and having a "modify user" scope felt maybe too general for just wanting to follow/unfollow, but I'm open to suggestions, lots of targeted scopes is not always the answer either
- Add `follow(Create|Update|Remove)` mutations that work similar to the existing buttons on the site
- Improved error handling for the resolver and lookup code I touched

Tested locally and attempted to write some tests

My first time writing Elixr so feedback would be very welcome. I'm not confident I did the right things to add support for the new scope specifically, I just grep'd for places the existing scope were referenced and it seems to work.

### DIFF
--- a/lib/glimesh/account_follows.ex
+++ b/lib/glimesh/account_follows.ex
@@ -9,6 +9,7 @@ defmodule Glimesh.AccountFollows do
   alias Glimesh.ChannelLookups
   alias Glimesh.Chat
   alias Glimesh.Repo
+  alias Glimesh.Streams.Channel
 
   def get_subscribe_topic(:follows), do: "accounts:follows"
   def get_subscribe_topic(:follows, streamer_id), do: "accounts:follows:#{streamer_id}"
@@ -28,27 +29,22 @@ defmodule Glimesh.AccountFollows do
   end
 
   def follow(%User{} = streamer, %User{} = user, live_notifications \\ false) do
-    changeset =
-      %Follower{
-        streamer: streamer,
-        user: user
-      }
-      |> Follower.changeset(%{
-        has_live_notifications: live_notifications
-      })
+    attrs = %{
+      has_live_notifications: live_notifications
+    }
 
-    case Repo.insert(changeset) do
+    {action, follower} =
+      case get_following(streamer, user, [:streamer, :user]) do
+        nil -> {:insert, %Follower{streamer: streamer, user: user}}
+        follower -> {:update, follower}
+      end
+
+    case follower |> Follower.changeset(attrs) |> Repo.insert_or_update() do
       {:ok, follower} ->
-        channel = ChannelLookups.get_channel_for_user(streamer)
-
         broadcast(follower, :followers)
 
-        if !is_nil(channel) and Glimesh.Chat.can_create_chat_message?(channel, user) and
-             sent_follow_message_recently?(channel, user) do
-          Chat.create_chat_message(user, channel, %{
-            message: " just followed the stream!",
-            is_followed_message: true
-          })
+        if action == :insert do
+          send_follow_message(streamer, user)
         end
 
         {:ok, follower}
@@ -79,8 +75,8 @@ defmodule Glimesh.AccountFollows do
     )
   end
 
-  def get_following(%User{} = streamer, %User{} = user) do
-    Repo.one(from f in Follower, where: f.streamer_id == ^streamer.id and f.user_id == ^user.id)
+  def get_following(%User{} = streamer, %User{} = user, preloads \\ []) do
+    Follower |> preload(^preloads) |> Repo.get_by(streamer_id: streamer.id, user_id: user.id)
   end
 
   def count_followers(%User{} = user) do
@@ -151,7 +147,20 @@ defmodule Glimesh.AccountFollows do
     )
   end
 
-  defp sent_follow_message_recently?(channel, user) do
+  defp send_follow_message(%User{} = streamer, %User{} = user) do
+    channel = ChannelLookups.get_channel_for_user(streamer)
+
+    if !is_nil(channel) and
+         Chat.can_create_chat_message?(channel, user) and
+         not_sent_follow_message_recently?(channel, user) do
+      Chat.create_chat_message(user, channel, %{
+        message: " just followed the stream!",
+        is_followed_message: true
+      })
+    end
+  end
+
+  defp not_sent_follow_message_recently?(%Channel{} = channel, %User{} = user) do
     follow_message = Chat.get_follow_chat_message_for_user(channel, user)
 
     time_since_last_follow_message =
--- a/lib/glimesh/account_follows.ex
+++ b/lib/glimesh/account_follows.ex
@@ -31,10 +31,11 @@ defmodule Glimesh.AccountFollows do
     changeset =
       %Follower{
         streamer: streamer,
-        user: user,
-        has_live_notifications: live_notifications
+        user: user
       }
-      |> Follower.changeset()
+      |> Follower.changeset(%{
+        has_live_notifications: live_notifications
+      })
 
     case Repo.insert(changeset) do
       {:ok, follower} ->
--- a/lib/glimesh/account_follows.ex
+++ b/lib/glimesh/account_follows.ex
@@ -28,33 +28,32 @@ defmodule Glimesh.AccountFollows do
   end
 
   def follow(%User{} = streamer, %User{} = user, live_notifications \\ false) do
-
-    result =
+    changeset =
       %Follower{
         streamer: streamer,
         user: user,
         has_live_notifications: live_notifications
       }
       |> Follower.changeset()
-      |> Repo.insert()
 
-    with {:ok, follower} <- result
-    do
-      channel = ChannelLookups.get_channel_for_user(streamer)
+    case Repo.insert(changeset) do
+      {:ok, follower} ->
+        channel = ChannelLookups.get_channel_for_user(streamer)
 
-      broadcast(follower, :followers)
+        broadcast(follower, :followers)
 
-      if !is_nil(channel) and Glimesh.Chat.can_create_chat_message?(channel, user) and
-          sent_follow_message_recently?(channel, user) do
-        Chat.create_chat_message(user, channel, %{
-          message: " just followed the stream!",
-          is_followed_message: true
-        })
-      end
+        if !is_nil(channel) and Glimesh.Chat.can_create_chat_message?(channel, user) and
+             sent_follow_message_recently?(channel, user) do
+          Chat.create_chat_message(user, channel, %{
+            message: " just followed the stream!",
+            is_followed_message: true
+          })
+        end
 
-      {:ok, follower}
-    else
-      {:error, error} -> {:error, error}
+        {:ok, follower}
+
+      error ->
+        error
     end
   end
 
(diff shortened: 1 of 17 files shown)


lib/glimesh/account_follows.ex
```diff
@@ -18,49 +18,52 @@ defmodule Glimesh.AccountFollows do
...
-        user: user
+        user: user,
+        has_live_notifications: live_notifications
       }
-      |> Follower.changeset(attrs)
+      |> Follower.changeset()
       |> Repo.insert()
 
-    channel = ChannelLookups.get_channel_for_user(streamer)
+    with {:ok, follower} <- result
```
danstiner, line 41: This `with` is to gracefully handle when `followChannel` mutation is invoked but the user is already following the streamer, previously the insert would through a constraint violation exception, now it jumps to the else and returns an error tuple that absinthe can display nicely.

lib/glimesh/account_follows.ex
```diff
@@ -18,49 +18,52 @@ defmodule Glimesh.AccountFollows do
...
-    results =
+    result =
       %Follower{
         streamer: streamer,
-        user: user
+        user: user,
+        has_live_notifications: live_notifications
       }
-      |> Follower.changeset(attrs)
+      |> Follower.changeset()
```
danstiner, line 38: Not a big deal but I think setting values when creating the object is cleaner and has better type checking than passing in attributes.

Converting the object to a changeset is needed to add constraint checks still, otherwise an insert with an existing row will throw.

lib/glimesh/api.ex
```diff
@@ -58,11 +59,13 @@ defmodule Glimesh.Api do
...
-        String.replace(acc, "%{#{key}}", to_string(value))
-      end)
-    end)
+    changeset
+      |> Ecto.Changeset.traverse_errors(fn {msg, opts} ->
+           Enum.reduce(opts, msg, fn {key, value}, acc ->
+             String.replace(acc, "%{#{key}}", to_string(value))
+           end)
+         end)
+      |> Enum.map(fn {key, value} -> "#{key}: #{value}" end)
```
danstiner, line 68: The errors on a changeset are a map, but afaict absinthe requires error values to be either a single item or a list. It would throw up big error messages at me when I was testing locally.

Mapping the map into a list of strings solves this

lib/glimesh/channel_lookups.ex
```diff
@@ -195,6 +195,10 @@ defmodule Glimesh.ChannelLookups do
     |> Repo.preload([:category, :user, :streamer])
   end
 
+  def get(id, preloads \\ []) do
```
danstiner, line 198: I had two issues with the `get_channel` method:
1. If there is no such channel, `get_by` will return nil which will throw when piped into the `preload` call
2. It always preloads the same associated rows, even when we don't needs them.

My experience with ORMs are shown the first most common performance issue is not preloading enough, the second most common is preloading too much. This approach both allows preloading exactly what you need and  handles missing rows gracefully. I did not replace the existing get_channel method to limit the scope of my changes.

lib/glimesh/api/resolvers/account_resolver.ex
```diff
@@ -29,6 +29,81 @@ defmodule Glimesh.Api.AccountResolver do
     end
   end
 
+  def follow_channel(
+        _parent,
+        %{channel_id: channel_id, live_notifications: live_notifications},
+        %{context: %{access: access}}
+      ) do
+    with :ok <- Bodyguard.permit(Glimesh.Api.Scopes, :follow, access),
```
danstiner, line 37: The way Elixr forces a single return from a method is interesting, had to read a bit about best patterns for handling failures. This is the cleanest approach I could find, any individual lookup can fail and the method will stop there and pass the error back to the caller, without throwing or having lots of nested if's.

The guards needed for the nil methods are kinda ugly. I think it would be cleaner if all the lookups returned ok/error result tuples, but that's a bigger change than I wanted to tackle. Also since all the error handling is the same it could potentially be extracted to a function. Or if needed, the individual lookups can be tagged so specialized errors can be generated for each like so:
```elixr
with :ok <- Bodyguard.permit(Glimesh.Api.Scopes, :follow, access),
     {:channel, channel} when channel != nil <- {:channel, Glimesh.ChannelLookups.get(channel_id, [:user])}
do
  # thing
else
  {:channel, nil} -> {:error, @error_no_such_channel}
end
```

lib/glimesh/account_follows.ex
```diff
@@ -18,49 +18,52 @@ defmodule Glimesh.AccountFollows do
...
-    results
+      {:ok, follower}
+    else
+      {:error, error} -> {:error, error}
+    end
   end
 
   def unfollow(%User{} = streamer, %User{} = user) do
-    Repo.get_by(Follower, streamer_id: streamer.id, user_id: user.id) |> Repo.delete()
+    if follower = Repo.get_by(Follower, streamer_id: streamer.id, user_id: user.id) do
```
danstiner, line 62: Similar story to `follow`, I needed this to gracefully handle there being no such follow and return an error tuple instead of throwing an exception (which results in the graphql endpoint returning a HTTP 500)

lib/glimesh/account_follows.ex
clone1018: The reason `changeset/2` is used is to actually validate that we allow changes to the property, and that it's being changed into a valid type. Take for example if `Follower` had `validate_subset(attrs, :has_live_notifications, [true, false])` inside it's changeset, then we could get the following benefit:
```elixir
iex(9)> %Glimesh.AccountFollows.Follower{has_live_notifications: "foobar"} |> Glimesh.AccountFollows.Follower.changeset(%{})
#Ecto.Changeset<
  action: nil,
  changes: %{},
  errors: [
    streamer: {"can't be blank", [validation: :required]},
    user: {"can't be blank", [validation: :required]}
  ],
  data: #Glimesh.AccountFollows.Follower<>,
  valid?: false
>

iex(11)> %Glimesh.AccountFollows.Follower{} |> Glimesh.AccountFollows.Follower.changeset(%{has_live_notifications: "foobar"})
#Ecto.Changeset<
  action: nil,
  changes: %{},
  errors: [
    streamer: {"can't be blank", [validation: :required]},
    user: {"can't be blank", [validation: :required]},
    has_live_notifications: {"is invalid", [type: :boolean, validation: :cast]}
  ],
  data: #Glimesh.AccountFollows.Follower<>,
  valid?: false
>
```

You're right though it has no benefit here, except for being able to be reused if for example a user has access to flip this raw data at some point.

lib/glimesh/api/resolvers/account_resolver.ex
```diff
@@ -29,6 +30,90 @@ defmodule Glimesh.Api.AccountResolver do
...
+         channel when channel != nil <- ChannelLookups.get(channel_id, [:user]),
+         streamer when streamer != nil <- Accounts.get_user(channel.user.id),
+         user when user != nil <- Accounts.get_user(access.user.id),
```
clone1018, line 41: You can convert these into pattern match guarantees, something like:
```suggestion
         %Channel{} = channel <- ChannelLookups.get(channel_id, [:user]),
         %User{} = streamer <- Accounts.get_user(channel.user.id),
         %User{} = user <- Accounts.get_user(access.user.id),
```

lib/glimesh/channel_lookups.ex
clone1018: Yeah, there's hundreds of cases where we preload data we likely don't even need.

danstiner: Probably ha, but it might not be a big impact, databases are fast!

My approach is just to "leave the code better than you found it" and I choose this as something to dig into and try to find a better pattern that could be used in the future.

lib/glimesh/api/resolvers/account_resolver.ex
danstiner: Perfect! Thanks I didn't learn this syntax from reading the pattern matching docs, updated the code and it's looking pretty clean at this point I think.

lib/glimesh/account_follows.ex
danstiner: Ah neat, I didn't realize that thanks. Switched back `changeset/2` as it was before.

lib/glimesh/api/resolvers/account_resolver.ex
```diff
@@ -29,6 +33,87 @@ defmodule Glimesh.Api.AccountResolver do
...
+         %User{} = user when user != nil <- Accounts.get_user(access.user.id) do
+      AccountFollows.unfollow(streamer, user)
+    else
+      {:ok, following} ->
+        {:ok, following}
+
+      {:error, %Ecto.Changeset{} = changeset} ->
+        {:error, Api.parse_ecto_changeset_errors(changeset)}
+
+      {:error, message} ->
```
danstiner, line 107: Removed the `is_binary` guard, it was preventing the `:unauthorized` atom returned from `Bodyguard` from being returned. absinthe calls to_string on non-binary values like atoms and handles them just fine.

A different approach would be converting `:unathorized` to a nicer error message, but that's a bigger change and having super pretty errors just isn't the most important thing to do imo, what we have is good enough for now.

test/glimesh/api/accounts_test.exs
```diff
@@ -138,6 +140,38 @@ defmodule Glimesh.Api.AccountsTest do
   }
   """
 
+  @follow_create_mutation """
+  mutation FollowCreate($channelId: ID!) {
+    followCreate(channelId: $channelId) {
```
danstiner, line 145: I changed how these mutations are named so they group together by the `follow` prefix. My prior experience with GraphQL is slowly coming back to me, this pattern of using prefixes to group related mutations is considered best practice as there is no other good way to indicate related mutations. It's not a perfect solution, but I think it's better than having `follow`, `unfollow`, `updateFollow` mutations.

Reference https://github.com/Shopify/graphql-design-tutorial/blob/master/TUTORIAL.md#naming-the-mutations

clone1018: Sorry for the delay, I understand the thought behind naming them this way, but it feels a bit counter intuitive given how verb based the rest of the API is `watchChannel`, `logStreamMetadata`, `uploadStreamThumbnail`. Instead of thinking abstractly in "fields that are updated" I like the API to be thought of as "actions that are preformed".

In this case, I propose having two mutations: 
* `follow(channelId: ID!, hasLiveNotifications: Boolean = false)` - Which is used for both following someone initially, and updating a follow to use live notifications or not. 
* `unfollow(channelId: ID!)` - Which is used to unfollow a channel.

clone1018: Let me know what you think, everything else in the PR is good with me!

danstiner: Thanks, 👍 perfectly happy sticking to verbs, updated the PR.

Agreed it's good to think of mutations as verbs, I try to do that also but just had a flashback to dealing with some very large APIs where that gets messy. Not so relevant with <20 mutations haha

There is only one edge case, and it's unlikely to ever come up. Calling `follow` and not specifying a value for `liveNotifications` course defaults to `false` for a new follow. The edge case is when there is an existing follow. I decided to have it change the value in the database to `false` in that case, vs just leaving it alone.

clone1018: Sounds good! Thanks for the work, one last review and I'll merge / deploy this morning.

PR #811, opened 2022-02-19
Adding ultrawide support for videos

mja00:
Currently when watching a stream on Glimesh while using an ultrawide(21:9 aspect ratio) it looks like this: 
image

Which cuts off a large portion of the bottom of the stream. This is due to using the `embed-responsive-16by9` class from Bootstrap which forces the container to be 16:9. 

This PR uses Phoenix's hooking system to send back info if the current user's browser is at least 21:9(a 2.3 ratio) and updates the DOM accordingly. This is done by calculating the ratio and if it's over a certain threshold it will fire an event to update the DOM with the `embed-responsive-21by9` class for the video container. This shouldn't fill up the websocket as it only fires an event if it absolutely needs to. I did try to do it fully client side but Phoenix kept hijacking the container and I really just didn't want to fight with it. 

Resulting in a stream that looks like this: 
image

While yes there's bars on the side of the video now, that's common with 16:9 video and basically unavoidable. 

Let me know if anything is wrong/funky, been a while since I touched any of this :P

### DIFF
--- a/assets/js/hooks/FtlVideo.js
+++ b/assets/js/hooks/FtlVideo.js
@@ -11,6 +11,21 @@ export default {
         let videoLoadingContainer = document.getElementById("video-loading-container");
         let forceMuted = container.dataset.muted;
         let saveVolumeChanges = false;
+        let currentlyInUltrawide = false;
+
+        // Handle 21:9 aspect ratio monitors/browers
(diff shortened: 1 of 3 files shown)


clone1018:
Merged with https://github.com/Glimesh/glimesh.tv/pull/822

PR #822, opened 2022-04-07
Fixed video container height for ultralarge viewports

cactysman:
This PR is based on #811, but adds simple CSS to ensure the video container height doesn't exceed the viewport height in whichever arbitrary viewport the user might be using.

**Before:**
image

**After:**
image

### DIFF
--- a/assets/css/glimesh/components/video.scss
+++ b/assets/css/glimesh/components/video.scss
@@ -58,3 +58,18 @@
 
     }
 }
+
+@media (min-width: 992px) {
+    #video-column {
+        max-height: calc(100vh - 90px);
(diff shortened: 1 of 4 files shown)


cactysman:
> Hey @cactysman thanks for submitting this, can you explain the difference between this and #811 for me? Should your changes also contain mja's changes?

Hey there, yes, that's why his commit is included here.

The difference is that #811 adds some support for 21:9 viewports, but the video player still grows larger than the page height (as demonstrated in the screenshots). So the CSS keeps it in place.

clone1018:
Looks good, thanks!

PR #832, opened 2022-05-08
Show recently used subcategories and tags per category on the stream info edit modal

MemoryLeakDeath:
### Feature Description
When a streamer edits their stream title and details from the stream page, we present them with two tag fields: the first field allows the streamer to specify their game/activity within the current category they have selected, the second field allows the streamer to add additional tags to further describe their stream to viewers and give prospective viewers keywords to help find their stream with.  

Currently we default the edit modal to the last category, sub-category (game/activity), and tags last saved by the streamer.  For streamers who switch activities/games and tags often, it can be tedious to delete/re-apply the appropriate activity and tags each time.  This feature aims to provide the streamer with a list of recently used sub-categories (games/activities) and tags from their previous streams in the currently selected category.  The lists will update appropriately as the streamer selects different categories from our category picker on the edit modal.

### Feature Requirements
- For the purposes of this feature, recent sub-categories and recent tags ONLY include sub-categories and tags used in previous streams.  It does not include sub-categories and tags saved without a corresponding stream.  For example, a streamer that has never streamed under the IRL category will not see recent sub-categories or tags (however, if the last saved category, sub-category, and tags were in the IRL category, those will still appear in the appropriate fields just as they do prior to this change).
- If the streamer has previously streamed in the currently selected category, display up to 5 of the most recent sub-categories they have used without duplicates (including not duplicating the last saved sub-category).
- If the streamer has previously streamed in the currently selected category, display up to 10 of the most recent tags they have used without duplicates (including not duplicating the last saved tags).
- A stream currently can only specify one sub-category (activity/game), thus selecting a recent sub-category should replace any existing sub-category in the field.
- A stream can have multiple tags, thus selecting a recent tag should append that tag to the end of the tags field.

### Feature Breakdown
- View of the edit stream details modal for a category the streamer has not streamed to and has no saved settings for:
recent-tags-no-stream-no-saved

- View of the edit stream details modal for a category the streamer has saved settings for but not streamed to:
recent-tags-no-stream-saved-settings

- View of the edit stream details modal for a category the streamer has streamed to previously but does not have saved settings for:
recent-tags-previous-stream-no-saved

- View of the edit stream details modal for a category the streamer has streamed to previously and has saved settings for:
recent-tags-previous-stream-saved-settings

### Additional Details
- The de-duplication process only considers the internal id of the tags, thus it is possible to have similar-looking tags suggested if those exist in the database with different ids in the current category.
- The recent sub-categories should be sorted from most recently used to least recent.
- The recent tags should be grouped by stream and presented in the same order they were used in the previous stream excluding any that duplicate currently saved tags.  The list is then sorted by most recent stream to least recent preserving the grouping.
- hovering over a recent sub-category or tag will show a tooltip with the number of uses of that tag site wide.  This can help to determine which tag to use when there are several similar worded tags.
- It was considered, but it is not currently possible to pre-populate the game/activity and tags fields with the tags used in the last stream conducted in that category.  Will consider this for a future patch if there is a desire expressed by the community.
- It is still possible to edit the stream category, sub-category, and tags from the "Channel Settings" page, however, no recent tags/sub-categories will be displayed there.

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -59,6 +59,7 @@ $fa-font-path: "/fa-fonts";
 @import "glimesh/components/apex-charts";
 @import "glimesh/components/video";
 @import "glimesh/components/channel-lookup-typeahead";
+@import "glimesh/components/recent-tags.scss";
 
 @import "glimesh/pages/dmca";
 @import "glimesh/pages/hosting.scss";
--- a/assets/css/glimesh/components/recent-tags.scss
+++ b/assets/css/glimesh/components/recent-tags.scss
@@ -1,0 +1,7 @@
+.recent-tag {
+    cursor: pointer;
+    font-size: medium;
+    margin-left: 0.5vh;
+    margin-right: 0.5vh;
+    margin-top: 1.25vh;
+}
--- a/assets/css/glimesh/variables.scss
+++ b/assets/css/glimesh/variables.scss
@@ -40,6 +40,8 @@ $color-teal: #48D1BD;
     --chat-bubble-background: #0e1726;
 
     --text-muted-color: #6c757d;
+
+    --hr-bg-color: #ffffff;
 }
 
 [data-theme="light"] {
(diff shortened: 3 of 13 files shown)


clone1018:
Sorry for the delay! I hope to have this reviewed for you very soon!

clone1018 (approved):

PR #835, opened 2022-05-12
Feature/community pride tags

Ribenaberri:
For Commuity Pride we wanted a way to make those who are supporting the event be easily identifiable. To that end we have added a new tag to the events page - Pride Event, which will tag it as separate to our usual community events.

unknown

unknown

(ignore the broken image link my PC is dumb sometimes)

And in addition to this anyone who uses the Community Pride tag as a channel descriptor will find it is Rainbow to match.

unknown

The color change will only happen for the Community Pride tag, all other tags remain unaffected.

***UPDATES***

Made suggested code changes from @clone1018 

Added Community Pride tag colors to show on live stream page and followers page

unknown

unknown

Also fixed issue with removal of blue color from the live streams page.

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -219,3 +219,10 @@ $fa-font-path: "/fa-fonts";
     -webkit-backdrop-filter: blur(12px) saturate(160%);
     background-blend-mode: exclusion;
 }
+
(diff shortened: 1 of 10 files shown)


clone1018:
Merge in #836!

PR #840, opened 2022-05-25
Preview profile markdown

aMytho:
This PR adds a markdown preview to the profile page.

The bootstrap tabs have been replaced by conditional rendering. The raw markdown data is passed to the server per keypress but is only converted when the user presses the preview button. The other form inputs have a phx-ignore attribute to prevent their value from being overwritten when the page changes.


clone1018:
Looks good! Works great locally too.

One question: any chance of switching this back to a tab like design? The buttons don't exactly suggest what is happening: 

image

aMytho:
I used the phoenix JS modules to remove/add the button backgrounds based on what is selected. It looks the same as the normal tabs now. The only difference is that preview uses `btn-info` instead of  `btn-primary`.

PR #841, opened 2022-05-29
Proposal: Add surface UI framework

clone1018:
Surface is a server-side rendering component library for Phoenix. Which in short means it provides a layer on top of Live View to add a more structured way of building components. 

Some of the main features are:
* Named / Typed props
* Multiple / Named Slots
* Reusable components

This PR adds initial support for it by converting the Events system over to it, which should nicely show off most of its functionality.


clone1018:
I've used this on another Phoenix project for a couple of months now, and I prefer it since it makes components much clearer. My only concern is that Live View is planning on eventually doing most of this natively, but the timeline is super unclear.

PR #852, opened 2022-06-21
Add Title Mutation to the API

aMytho:
This PR adds the ability to set a title from the API. Useful for third party devs and the mobile app.

If a title is too short (empty string) the default title is set. If the title is too long, the channel doesn't exist, or the user doesn't have permission the user is sent a message informing them of the error.

I also added a new scope `title`.  The scope works fine but I'm having issues getting the project to recognize the new scope. I added the scope to the `oauth_migration` file and did a DB reset which solved the problem. I don't know if you can run that migration file again or If I need to write a new one. If there is something I need to change let me know!


lib/glimesh/api/resolvers/channel_resolver.ex
```diff
@@ -31,6 +32,23 @@ defmodule Glimesh.Api.ChannelResolver do
...
 
+  def change_title(_parent, %{channel_id: channel_id, title: title},
+        %{context: %{access: access}
+      }) do
+    with :ok <- Bodyguard.permit(Glimesh.Api.Scopes, :title, access) do
+      channel = Glimesh.ChannelLookups.get_channel(channel_id)
+
+      if String.length(title) > 250 do
+        {:error, "Title must less than 250 characters."}
+      else
```
clone1018, line 43: This validation should exist in the changeset already, I don't think we should duplicate it. You can listen for the {:error, changeset} and use a helper to spit it back out to the client.

Example: https://github.com/Glimesh/glimesh.tv/blob/dev/lib/glimesh/api/resolvers/account_resolver.ex#L38-L44

lib/glimesh/streams/title.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Glimesh.Streams.Title do
+  alias Glimesh.Repo
+  alias Glimesh.Streams.Channel
+  alias Glimesh.Events
+
+  def change_title(channel, title) do
```
clone1018, line 6: This should already be handled by `Glimesh.Streams.update_channel(streamer, channel, %{title: "some title"})`

clone1018 (changes requested):
Great feature! Just a couple of changes we should make, and I'd like your thoughts on converting the scope to be more encompassing of all the Stream Info you can change via the website. We don't have to add them all to the API now though

lib/glimesh/api/schema/channel_types.ex
```diff
@@ -106,6 +106,14 @@ defmodule Glimesh.Api.ChannelTypes do
 
       resolve(&ChannelResolver.upload_stream_thumbnail/3)
     end
+
+    @desc "Update a channel's title"
+    field :change_title, type: :channel do
+      arg(:channel_id, non_null(:id))
+      arg(:title, non_null(:string))
+
+      resolve(&ChannelResolver.change_title/3)
```
clone1018, line 115: In respect to the other comment I made, this could be `update_channel` / `update_channel_info` or something similar, to allow us easier access to add new changeable fields.

aMytho:
I switched the `title` scope to `stream_info`. The functions have been modified as well. `changeTitle` mutation is now `updateStreamInfo` but it currently only accepts a title argument. We can modify that later to add more fields.

For now `stream_info` scope can only modify the title. Later we need to decide which fields it is allowed to change. I think the basic stream fields (Title, Category, Subcategory, Tags) are fine. However, there are some other fields that users could change per stream that probably shouldn't be in that scope (mature setting, language, etc). For those fields we will need to make a new scope `channel_settings` or similar. I prefer things to be as fine grained as possible (least privilege) so we may want to add a few scopes to match what is on the channel settings page (`channel_customization`, `stream_information`(we have this now), `chat_customization`, `homepage`). That can be decided later though.

clone1018:
Totally agree with you on granular controls. I think Stream Info is the first logical grouping (title, category, subcategory, tags), and then as you said, as we add more to the API we can add additional scopes as necessary. For the GQL side though we can limit the params based on the scope if necessary.

lib/glimesh/api/resolvers/channel_resolver.ex
```diff
@@ -31,6 +31,26 @@ defmodule Glimesh.Api.ChannelResolver do
     end
   end
 
+  def update_stream_info(_parent, %{channel_id: channel_id, title: title},
+        %{context: %{access: access}
+      }) do
+    with :ok <- Bodyguard.permit(Glimesh.Api.Scopes, :stream_info, access) do
+      channel = Glimesh.ChannelLookups.get_channel(channel_id)
+      if channel !== nil do
```
clone1018, line 39: No reason to change this PR, but in the future you can use `with` for multiple checks. For example:
```
with :ok <- Bodyguard.permit(Glimesh.Api.Scopes, :stream_info, access),
      %Channel{} = channel <- Glimesh.ChannelLookups.get_channel(channel_id), 
      {:ok, changeset} <- Streams.update_channel(access.user, channel, %{title: title})  do
        # Your Happy Path
else
    # Any unhappy paths
end
```

aMytho: Alright. Is there anything else I need to change?

clone1018: No sorry, just need to find time personally to merge & deploy!

PR #858, opened 2022-08-14
Country Updates

Heiwa1580:
This change is to update our list of countries to the most recent Stripe list. There are two changes with this list.

1.) I've removed all the Stripe preview program countries. Stripe at any point can stop supporting these countries if they need to fix something. This happened to us with Brazil. I think its best we make sure countries are fully finished with Stripe and are not preview program countries.

2.) Many countries have moved out of the preview countries and are now fully supported. I've added these countries to the list.


clone1018:
Sorry for the delay on this, I want to check out tomorrow how exactly these countries are removed, and figure out if we have any users onboarded inside them.

clone1018:
Completed in https://github.com/Glimesh/glimesh.tv/pull/871 !

PR #860, opened 2022-08-24
adding png for emotes

Ribenaberri:
So i'm not sure if this was the best way to code things but it was the best way I could figure to make it work. 

Emotes now support PNG! 

PNG would be the only way for users to upload emotes going forward - this would enable more emotes to be uploaded by the user. 

Anything SVG should be retained, this will only effect emotes going forward for channels. 

Platform Emotes are still SVG as they have always been. 

My dev environment allowed more than 2 emotes to be uploaded, so that part may need checking (it was set at 10, not something I had set myself, upped to 100 now)

Any issues need addressing let me know. 

Not really much to show screenshot wise :) 

Thanks.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -273,7 +273,7 @@ config :glimesh, Glimesh.Socials.Twitter,
   access_token: "",
   access_token_secret: ""
 
-config :glimesh, Glimesh.Emotes, max_channel_emotes: 10, allow_channel_animated_emotes: true
+config :glimesh, Glimesh.Emotes, max_channel_emotes: 100, allow_channel_animated_emotes: true
 
 config :libcluster,
   topologies: []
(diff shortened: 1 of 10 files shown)


haydenmc:
> PNG would be the only way for users to upload emotes going forward - this would enable more emotes to be uploaded by the user.

I'm a little bit concerned about regressing the ability to allow users to upload vector graphics for their emotes.

Is it too costly to provide support for both raster and vector images?

Ribenaberri:
> 

Its because there's later framework coming which separates out stickers which will be high quality files and emotes which will be lower quality I.e. SVG. Also right now the majority of users don't use native SVG files, rather they convert them from PNG. This is a medium term measure designed to help the user base upload more emotes easier and allow us yo start getting framework in place for the larger vision of stickers/emotes as separate entities.

PR #861, opened 2022-08-25
PNG support for Emotes

Ribenaberri:
So i'm not sure if this was the best way to code things but it was the best way I could figure to make it work.

Emotes now support PNG!

PNG would be the only way for users to upload emotes going forward - this would enable more emotes to be uploaded by the user.

Anything SVG should be retained, this will only effect emotes going forward for channels.

Platform Emotes are still SVG as they have always been.

My dev environment allowed more than 2 emotes to be uploaded, so that part may need checking (it was set at 10, not something I had set myself, upped to 100 now) **edit, have since changed that back to 10**

Any issues need addressing let me know.

Not really much to show screenshot wise :)

Thanks.

**This is a new branch after i made some mess ups, this is now WORKING (Cheer) Full write up on what this is and why it is will be available on building glimesh**


clone1018:
Thanks & apologies for the delays!

PR #862, opened 2022-09-03
Giving users the power to turn off receipt of Gift Subs

Ribenaberri:
Small QoL Update after a conversation in discord recently. 

Giving users the power to turn off receiving gift subscriptions. Some users prefer not to bad able to receive them, this could be a host of reasons as to why a user may not want to receive gift subs. With this in mind, and inkeeping with the Glimesh ethos of giving everyone a choice over how their accounts can function, I have now added this capacity. 

Located under user preferences is a toggle, this is on by default but users may turn it off if they wish. 

unknown1

unknown

If this is toggled off any user trying to give that person a gift sub will be notified by an error message before proceeding to checkout, which will prevent the person gifting from being charged. 

Further discussion will be available in a Building Glimesh thread. 

Thanks all!

### DIFF
--- a/lib/glimesh/accounts/user_preference.ex
+++ b/lib/glimesh/accounts/user_preference.ex
@@ -13,6 +13,7 @@ defmodule Glimesh.Accounts.UserPreference do
     field :show_timestamps, :boolean, default: false
     field :show_mod_icons, :boolean, default: true
     field :show_mature_content, :boolean, default: false
+    field :gift_subs_enabled, :boolean, default: true
 
     timestamps()
   end
@@ -27,7 +28,8 @@ defmodule Glimesh.Accounts.UserPreference do
       :site_theme,
       :show_timestamps,
       :show_mature_content,
-      :show_mod_icons
+      :show_mod_icons,
+      :gift_subs_enabled
     ])
     |> unique_constraint(:user_id)
   end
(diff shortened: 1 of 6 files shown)


clone1018:
Thanks for the PR! I'd actually rather have this implemented more similar to how we have the other support modal options. They are stored in the `channel` table, under names like: `show_subscribe_button`, `show_streamloots_button`, etc. Can you pivot this PR to use that table instead and update the UI to be in the Channel Addons tab? https://glimesh.tv/users/settings/addons

Here's how most of it is wired up: https://github.com/Glimesh/glimesh.tv/search?q=show_donate_button

Ribenaberri:
> Can you pivot this PR to use that table instead and update the UI to be in the Channel Addons tab? https://glimesh.tv/users/settings/addons

That would only work if a user has a channel though for it to be visible, this isn't on a channel level. This is so all individuals can have the power to turn off the capacity to be gifted a subscription by other users. Being in the channel addons tab would not be a good fit for this as its for the individual person, streamer, or viewer, to be able to use. 

We can of course also add this in so a streamer an turn off all gift subs for their channel, but that is different to what this particular PR is about.

clone1018:
FWIW my comment above is useless as I didn't understand this was to turn off gift sub receiving... :D

PR #879, opened 2022-11-06
Added optional moderator edit permissions to allow editing stream title, category, and tags.

MemoryLeakDeath:
# New edit channel title, category, and tags permission for moderators
Streamers can now give edit permissions to individual moderators on their channels.  The edit permission will give the selected moderators access to the edit modal on the stream page.  It will not give moderators permission to any of the streamer's settings pages.  The permission is off by default.

image

image

## Stream page update
Moderators with edit permission will now be able to access the edit modal on the stream page:

image

image

## Moderation Log update
Whenever a moderator with edit permissions makes changes to the stream title/category/tags an additional log message will appear in the moderation log:

image

## Database Updates
- Added "is_editor" boolean column to channel_moderators table
- Modified the "user_id" column to allow null on the channel_moderation_log table

## Code Flow Update
The edit stream modal now goes though a separate changeset from the update channel changeset so it can be isolated permission wise.


clone1018:
If you can, rebase this against `dev` (the main branch) and the tests should magically pass.

MemoryLeakDeath:
Rebased and it works -- thanks for looking into this!

clone1018:
Looks good, thanks!

PR #882, opened 2022-11-16
Fix typos

kianmeng:
Found via `codespell -S priv,deps,_build,*.svg -L nd,ro,nam,wth,ue,fof,cliente,beastiality,  extacy,masterbation,shat,toke`


clone1018:
Hey there, thanks for this! A couple of the tests that rely on these misspelt words, so they are not failing. Could you fix them? Maybe running codespell without -S test would be enough.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #885, opened 2022-11-21
Prompt streamers when they are eligible to be shown on the homepage

MemoryLeakDeath:
# Description
We provide streamers with the option of potentially being featured on the homepage after they have streamed at least ten hours on the platform.  It has been suggested on the building Glimesh forum that we remind streamers when they are eligible for homepage placement as it appears many new streamers may have missed that setting on the stream settings page.  To that end, I will display the following banner to the channel owner only on their channel page:

image

The above banner will only display if the following conditions are met:
- The person viewing the channel page is the channel owner
- The channel owner has streamed at least 10 hours
- The channel owner has not already opted in to being displayed on the homepage
- The channel owner has not clicked the "Dismiss" button on the banner

## Notification banner buttons
The notification banner has a "Go to Settings" and a "Dismiss" button.  The "Go to Settings" button will redirect the owner to the stream settings page which contains the homepage opt-in button at the bottom.  The "Dismiss" button will set a flag on the channel to dismiss the notification and never prompt about it again.

# Technical Details
## New background job
I've created a new background Oban job called "prompt_homepage_opt_in" that is set to run every 30 minutes and will accomplish the following tasks:
- Find all channels that are accessible (inaccessible = false) and are opted in to be displayed on the homepage (show_on_homepage = true) then set "prompt_for_homepage" to "ignore" so those channel owners will not be prompted as they are already opted in.
- Find all channels that are accessible (inaccessible = false), are NOT opted in to be displayed on the homepage (show_on_homepage = false), and have at least 10 hours of streaming time then set "prompt_for_homepage" to "prompt".  The owners of those channels will see the notification banner the next time they go to their channel page.

## New column and enumeration
I've added the column "prompt_for_homepage" that defaults to "ineligible" to the channels table.  It refers to a defined database enumeration called "homepage_prompt_type" which can take on one of the following values: "ineligible", "prompt", or "ignore".  The column is only set to "prompt" by the background job.  The UI will only update the column to "ignore".

### DIFF
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -256,7 +256,8 @@ config :glimesh, Oban,
        {"* * * * *", Glimesh.Jobs.StreamMetricsCron},
        {"*/10 * * * *", Glimesh.Jobs.AutoHostCron},
        {"*/5 * * * *", Glimesh.Jobs.HomepageCron},
-       {"*/5 * * * *", Glimesh.Jobs.StreamPrunerCron}
+       {"*/5 * * * *", Glimesh.Jobs.StreamPrunerCron},
+       {"*/30 * * * *", Glimesh.Jobs.PromptHomepageOptIn}
      ]}
   ]
 
--- a/lib/glimesh/channel_lookups.ex
+++ b/lib/glimesh/channel_lookups.ex
@@ -301,4 +301,44 @@ defmodule Glimesh.ChannelLookups do
       []
     end
   end
+
+  # This checks to see if a channel has met the criteria for showing on the homepage
+  # but the channel owner has not yet opted in.
+  def update_prompt_channel_opt_in_for_homepage do
+    ten_hour_query =
+      from s in Glimesh.Streams.Stream,
+        select: s.channel_id,
+        group_by: s.channel_id,
+        having: sum(s.ended_at - s.started_at) >= fragment("INTERVAL '10 hours'")
+
+    update_query =
+      from(c in Channel,
+        where: c.inaccessible == false,
+        where: c.show_on_homepage == false,
+        where: c.prompt_for_homepage == :ineligible,
+        where: c.id in subquery(ten_hour_query)
+      )
+
+    Repo.update_all(update_query, set: [prompt_for_homepage: :prompt])
+  end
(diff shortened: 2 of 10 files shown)


MemoryLeakDeath:
Looks like there's a transient postgres error here.  Hopefully a re-run of the tests will clear it.

Cykotiq:
That feels a bit aggressive and very much in-your-face. I feel like the standard bootstrap alert with a couple added buttons might look a little cleaner
image

("Go To Settings" and "Dismiss" is probably better wording, but you get the idea)

Ribenaberri:
> That feels a bit aggressive and very much in-your-face. I feel like the standard bootstrap alert with a couple added buttons might look a little cleaner image
> 
> ("Go To Settings" and "Dismiss" is probably better wording, but you get the idea)

I feel like what Cykotiq has here is much more in line with how Glimesh currently operates, the color schemes it uses and the button layouts :)

MemoryLeakDeath:
@Cykotiq @Ribenaberri 

I'll bring the colors more in-line with our info boxes as you've shown.  As far as the GCT suspension checks, I believe we do the same checks as the homepage job, but I'll verify that.

Ribenaberri:
> @Cykotiq @Ribenaberri
> 
> I'll bring the colors more in-line with our info boxes as you've shown. As far as the GCT suspension checks, I believe we do the same checks as the homepage job, but I'll verify that.

Im not entirely sure how those homepage checks work presently, the GCT system needs an overhaul as at present I don't think it itself "saves" the infraction history - it may be we need some additional work on the GCT side before we can do this so that it tracks fully to come in alongside this feature!

Thanks for checking :)

MemoryLeakDeath:
@Cykotiq @Ribenaberri 

I've updated the banner to take up less space and be a little less harsh on the eyes (as much as bootstrap will allow at least).

image

Ribenaberri:
Looks muchhhh better now more in line with the general way things look on Glimesh, thanks for that!

clone1018:
Thanks for the contribution! Going to hold off on this until our new notifications system is available.

PR #889, opened 2022-11-27
Highlight new streamers with a badge and filter option

MemoryLeakDeath:
It may be beneficial to community-building if we can quickly identify new streamers on the platform and welcome them to Glimesh. To this end, we can add a special “tag” that can be selected from the filters at the top of category pages. We can also put a special badge next to the streamer’s tile to indicate that they are new to Glimesh.

## New filter option on category pages
At the top of the category pages viewers will now be able to filter the live streams to show only those from new streamers on the platform:
image

image

Clicking the toggle button will apply the new streamer filter to any existing filters already active.

## New badge on category pages, homepage, and following page
Streamers flagged as new on the platform will have a red badge that will appear on their stream panel next to their language/mature badges as shown below:
image
image

## How we determine if a streamer is new
I've introduced a new background job called "is_new_streamer_flagger" that runs once per hour and will use the following criteria to determine if a channel qualifies as new:
- is it within 7 days of the channel's first stream (as determined by the stream start time)
- does the channel have 5 or fewer streams (this is to not unfairly punish streamers who can only stream a few times a week or who run a test stream then start streaming the next week).
- the channel must NOT have greater than 50 total hours streamed (this is to prevent 24/7 streams from having the new badge well after their initial period).

## Development details
- A new boolean column "is_new_streamer" has been added to the channels table.
- The new field should only be written to from the new "is_new_streamer_flagger" job.
- The API should NOT have access to this new channel field as it could allow gaming the system.
- I've added a new job called "is_new_streamer_flagger" which will run every hour and perform updates against the new column for channels that are not flagged as inaccessible (inaccessible = false).
- UI only reads the new column.
- Added filter option to the category pages.

### DIFF
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -256,7 +256,8 @@ config :glimesh, Oban,
        {"* * * * *", Glimesh.Jobs.StreamMetricsCron},
        {"*/10 * * * *", Glimesh.Jobs.AutoHostCron},
        {"*/5 * * * *", Glimesh.Jobs.HomepageCron},
-       {"*/5 * * * *", Glimesh.Jobs.StreamPrunerCron}
+       {"*/5 * * * *", Glimesh.Jobs.StreamPrunerCron},
+       {"*/60 * * * *", Glimesh.Jobs.IsNewStreamerFlagger}
      ]}
   ]
 
--- a/lib/glimesh/channel_lookups.ex
+++ b/lib/glimesh/channel_lookups.ex
@@ -8,7 +8,7 @@ defmodule Glimesh.ChannelLookups do
   alias Glimesh.AccountFollows.Follower
   alias Glimesh.Accounts.User
   alias Glimesh.Repo
-  alias Glimesh.Streams.{Category, Channel, ChannelHosts}
+  alias Glimesh.Streams.{Category, Channel, ChannelHosts, Stream}
 
   ## Filtering
   @spec search_live_channels(map) :: list
@@ -20,6 +20,7 @@ defmodule Glimesh.ChannelLookups do
     |> apply_filter(:subcategory, params)
     |> apply_filter(:tags, params)
     |> apply_filter(:language, params)
+    |> apply_filter(:is_new_streamer, params)
     |> order_by(fragment("RANDOM()"))
     |> group_by([c], c.id)
     |> preload([:category, :subcategory, :user, :stream, :tags])
@@ -62,6 +63,15 @@ defmodule Glimesh.ChannelLookups do
 
   defp apply_filter(query, :language, %{"language" => languages}) when is_list(languages) do
     where(query, [c], c.language in ^languages)
+  end
+
+  defp apply_filter(query, :is_new_streamer, %{"is_new_streamer" => is_new_streamer})
+       when not is_nil(is_new_streamer) do
+    if is_new_streamer == "true" do
+      where(query, [c], c.is_new_streamer == ^is_new_streamer)
+    else
+      query
+    end
   end
 
   defp apply_filter(query, _field, _params) do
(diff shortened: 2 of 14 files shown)


Ribenaberri:
Is there a way for a new streamer to opt out of this, is this a consideration worth having? Not every new streamer may want to draw attention to themselves in such an obvious way. They could be nervous, or be wanting to test things out in the early stages without too much attention being drawn too them. 

Some kind of opt-out (i do believe opt-in should be default) at point of channel creation would be a good consideration here.

MemoryLeakDeath:
@Ribenaberri 
In my opinion, having an opt-out defeats the purpose of it.  I understand they may be testing stuff, but putting "test stream" as their title should be sufficient I feel.  Nervousness is understandable, but I don't think there will be a flood of people overwhelming them.

MemoryLeakDeath:
rebased with dev.

Ribenaberri:
> but I don't think there will be a flood of people overwhelming them.

While this may be the case at the present moment, certainly longer term there may be. 

While its not especially a hill I'm willing to die on I do want to take the time to fully explain my thoughts on this feature (I do my best thinking after walking the dog so can better articulate than I could yesterday) and i do have a couple of additional questions more relating to how it would work at a broader thought from my experience here :) 

- We offer up full control and freedom for how a user describes their content on the platform using the tags (so long as it remains within the ToS & RoC) making this tag a default "does exist" removes some of that control over that from the user

- New streamers if we make them super obvious could become an easier target for bad actors, as they could be seen as easier targets more likely to be targeted due to less chance of moderators etc, we shouldn't be leading anyone in the community to need to feel like they need to have their mod settings ramped up full before they even start to stream with us - which is honestly where a lot of my thoughts and concerns on this stem from, as this may genuinely occur to some people who are newer or just simply don't want much attention being brought to themselves because of the background they came from if they came from an ethos of hate raids or what have you. 

- How will this handle accounts which get deleted, fully deleted I mean, but come back a few months later? I speak this from experience as there are many users to the platform who rotate between coming to check us out, deleting their accounts entirely, then coming back a few months later - obviously because the user has deleted their account, when they come back they make an entirely new account with a new userID etc attached too it. This could lead to a handful of streamers seemingly always being tagged as new (especially in our current smaller state). 

- I assume if someone deactivated their channel during this "new tag" process and then reactivated it the time left for the tag would still be the there and just continue from where it last left off? Much like the cumulative totals for the homepage feature>

MemoryLeakDeath:
> * We offer up full control and freedom for how a user describes their content on the platform using the tags (so long as it remains within the ToS & RoC) making this tag a default "does exist" removes some of that control over that from the user

Granted.  But we're not doing anything that other services also don't do.  Discord for instance shows a sprout icon next to new users that I don't think they can control.  And again, it's a temporary status, after the qualifications above expire the tag is no longer available.

> * New streamers if we make them super obvious could become an easier target for bad actors, as they could be seen as easier targets more likely to be targeted due to less chance of moderators etc, we shouldn't be leading anyone in the community to need to feel like they need to have their mod settings ramped up full before they even start to stream with us - which is honestly where a lot of my thoughts and concerns on this stem from, as this may genuinely occur to some people who are newer or just simply don't want much attention being brought to themselves because of the background they came from if they came from an ethos of hate raids or what have you.

In this case, I feel like the benefits outweigh the drawbacks.  Bad actors will always exploit whatever angle they can, I have not exposed this filter to the API, so they will have to navigate the site to find newbies.  Also the tag can be utilized by the GCT to find new users and maybe lurk in their streams to help out with questions or to provide a measure of protection against bad actors.

> * How will this handle accounts which get deleted, fully deleted I mean, but come back a few months later? I speak this from experience as there are many users to the platform who rotate between coming to check us out, deleting their accounts entirely, then coming back a few months later - obviously because the user has deleted their account, when they come back they make an entirely new account with a new userID etc attached too it. This could lead to a handful of streamers seemingly always being tagged as new (especially in our current smaller state).

That is an exploit -- but I feel that is more of a social engineering problem then a software one.  I can't do much to stop someone from creating account1 then account2, etc.  It would be self-defeating since each account would start out with zero followers and have to go through the setup to have subscribers or to get emotes approved, etc.  I wouldn't put it past someone to try it, but I doubt it would be very effective.

> * I assume if someone deactivated their channel during this "new tag" process and then reactivated it the time left for the tag would still be the there and just continue from where it last left off? Much like the cumulative totals for the homepage feature>

Correct.  The timer is not based on account lifetime, but on the start date of your first stream.  A user could have an account for a year without streaming and then still be considered new on their first stream.

PR #891, opened 2022-12-03
Add a share button to the channel page ("Sharing is Caring")

MemoryLeakDeath:
I've added a share button to the bottom of the stream panel next to the debug and report user buttons.  I've re-styled the debug and report user buttons so all three buttons align properly.  The debug and report user buttons should continue to function the same as they have prior to this change (the debug icon should still change to yellow when there are stream connection issues).

image

The new share button is available to both logged-in and anon users.  Clicking the button will open the following modal dialog box:
image

The various buttons on the modal will open the corresponding social media sites in a new browser tab and place the user (if they are logged in on the social media site of choice) onto a page where they can edit/create a post.  
image

Streamers may set the default post message from their channel settings page:
image

Viewers may edit the message before they submit the post on the various social media sites -- the default message is just a suggestion.  In addition to the message, the feature will automatically add a link to the streamer's channel page.  The system default message is: "Come and enjoy this #Glimesh stream with me!".  The maximum size of this message is restricted to 100 characters on the user interface.

## Development Notes
- Added column "share_text" to the "channels" table.  This will contain the default text that will appear when sharing the stream.
- Added a MastodonShareButton javascript hook to handle the creation of the share link because mastodon requires the instance name to be entered for the correct share panel to appear.


MemoryLeakDeath:
rebased with dev

PR #893, opened 2022-12-11
Homepage redesign and Live stream category counts

MemoryLeakDeath:
I have reorganized the homepage layout so highlighted streams appear above the fold and our categories show above the streams.  In addition, I've added the count of live streams per category under the category buttons so viewers are more aware that there are more then just 6 streams live on the site on any given time.  I've also removed most of the "marketing" text on the homepage as it detracts from the user content and is not relevant to returning viewers.

## Homepage with Community Event and spotlighted streams
image

## Homepage without an event but with spotlighted streams
image

## Homepage with no spotlighted streams but user logged in
image

## Homepage with no spotlighted streams and user NOT logged in
image

### DIFF
--- a/lib/glimesh/channel_lookups.ex
+++ b/lib/glimesh/channel_lookups.ex
@@ -301,4 +301,16 @@ defmodule Glimesh.ChannelLookups do
       []
     end
   end
+
+  def count_live_channels_by_category do
+    Repo.replica().all(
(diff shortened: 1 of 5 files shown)


clone1018:
With the beta redesign, this PR will not be needed in its current form. Thank you for proposing the idea!

PR #894, opened 2022-12-20
Show random live subcategories and tags on homepage

MemoryLeakDeath:
## Description
To help surface more live streams on the homepage and introduce users to the category search function, I've added a tag cloud on the homepage.  The cloud will be a random mix of both subcategories and tags taken from currently live channels across the platform regardless of time streamed.

image

Clicking a tag will automatically take the user to the appropriate category search page with the subcategory or tag already selected in the filter at the top of the page:

image

image

## Limitations
- A maximum of 20 subcategories/tags will show on the homepage.
- Despite my best efforts, the database query tends to favor some categories over others when it comes to removing duplicate tags.  For example, having two live streams -- one in the Art category and one in the Gaming category -- that both use the "Community Pride" tag will predominately surface one category over the other thus showing only one of the streams.  This could be solved by having a more generic tag search that spans across categories.

### DIFF
--- a/lib/glimesh/channel_lookups.ex
+++ b/lib/glimesh/channel_lookups.ex
@@ -8,7 +8,7 @@ defmodule Glimesh.ChannelLookups do
   alias Glimesh.AccountFollows.Follower
   alias Glimesh.Accounts.User
   alias Glimesh.Repo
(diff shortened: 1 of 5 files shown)


clone1018:
With the beta redesign, this PR will not be needed in its current form. Thank you for proposing the idea!

PR #899, opened 2023-01-06
Various wording and bug fixes

Ribenaberri:
Few things that have been piling up - 

FIXED: 

- The delete button for the events team dashboard
- Duplicate emotes showing for subscribers when an emote is made available globally (this is a separate fix to the previous fix, which only fixed them locally)
- The link on the FAQs on the main site that linked to the Glimesh Discord

CORRECTED: 

- The links on the stream setup page to the support pages, after the support revamp these links no longer worked
- FAQs on the main site had a section where a sentence was repeated twice, this is also now corrected

### DIFF
--- a/lib/glimesh/events_team.ex
+++ b/lib/glimesh/events_team.ex
@@ -111,7 +111,7 @@ defmodule Glimesh.EventsTeam do
   def delete_event(eventid) do
     %Event{id: String.to_integer(eventid)}
     |> Event.changeset()
-    |> Repo.delete()
+    |> Repo.delete(eventid)
   end
 
   def list_types do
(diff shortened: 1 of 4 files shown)


lib/glimesh/events_team.ex
```diff
@@ -111,7 +111,7 @@ defmodule Glimesh.EventsTeam do
   def delete_event(eventid) do
     %Event{id: String.to_integer(eventid)}
     |> Event.changeset()
-    |> Repo.delete()
+    |> Repo.delete(eventid)
```
clone1018, line 114: I don't think this change does what you expect, the documents say this param is for options, not IDs.

Ribenaberri: I was trying to fix the delete button because it wasn't working, when I tested it did delete the event now.

PR #904, opened 2023-02-25
Only call Twitter API when necessary

### LINKED ISSUE #634: Don't call the Twitter API until the user requests it
Currently we poke the Twitter API every single time someone requests any of the pages under UserSettingsController. 

https://github.com/Glimesh/glimesh.tv/blob/dev/lib/glimesh_web/controllers/user_settings_controller.ex#L136

Ideally we should hide expensive requests like this behind an initial user opt in to generate the link. The same would apply for TaxIDPro Form Link & Stripe Account Link.

MemoryLeakDeath:
fixes #634 

Currently we are calling the Twitter API on each settings page load unnecessarily.  I've removed the calls to the api and consolidated it to a new endpoint on the UserSocialController -- "/users/social/twitter/connect".

The new endpoint will do the Twitter authorization_url lookup and, if successful, do the redirect to Twitter that the "Connect with Twitter" button on the user profile page would have done.  If we are not successful in contacting the Twitter API, the endpoint will redirect the user back to the user profile page with an error message:

image

NOTE: I am awaiting access from Twitter to their API, so I am only able to test the error path and not the successful path at this time.

There should be no other changes to the Twitter auth flow as a result of this change.

### DIFF
--- a/lib/glimesh_web/controllers/gct_controller.ex
+++ b/lib/glimesh_web/controllers/gct_controller.ex
@@ -116,7 +116,6 @@ defmodule GlimeshWeb.GctController do
   def edit_user_profile(conn, %{"username" => username}) do
     current_user = conn.assigns.current_user
     user = Accounts.get_by_username(username, true)
-    twitter_auth_url = Glimesh.Socials.Twitter.authorize_url(conn)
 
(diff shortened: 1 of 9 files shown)


clone1018:
Merged in https://github.com/Glimesh/glimesh.tv/commit/55812ed021b6b20c99950e7770705763cceb44c5

PR #909, opened 2023-04-04
Fix URL by removing duplicate forward slash

tiagopinhotx:
Previously, the URL was broken due to an extra forward slash in the path. This pull request removes the duplicate forward slash and fixes the link on the website.


clone1018:
Awesome, thanks!

PR #910, opened 2023-04-08
Fix route to add mods by removing /create

tiagopinhotx:
Fix bug where route to add mods (`"/users/settings/channel/mods/create"`) was returning not found. Removing the /create should fix this issue since the :create route is on the mods route itself. 

Routes Screenshot
image


clone1018 (approved):
Perfect fix!

